PR #1, opened 2023-03-07
chore: organizing docs to use expected directories

heckj:
- updated build script to match


scripts/build-ghpages-docs.sh
```diff
@@ -6,17 +6,24 @@ THIS_SCRIPT_DIR="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
...
-    --fallback-bundle-version 1 \
-    --additional-symbol-graph-dir $BUILD_DIR \
+    --output-path ${PACKAGE_PATH}/docs \
+    --emit-digest \
+    --disable-indexing \
     --transform-for-static-hosting \
-    --hosting-base-path automerge-swifter \
-    --output-path $PACKAGE_PATH/docs/
+    --hosting-base-path 'automerge-swifter' \
+    --source-service github \
```
heckj: NOTE: the last 3 lines on here are new/different to the Github docs you'd set up previously. When built with Swift 5.8, they'll add in links to the GitHub source to the existing documentation. I'm not sure if you'd like that or not, but I went ahead and left it in from replicating work in another project. Can easily be stripped out if y'all would prefer.

alexjg: Interesting, I get this error when running the script:

```
Error: Unknown option '--source-service'
Usage: docc convert [<options>] <source-bundle-path>
  See 'docc convert --help' for more information.
  ```
  
 I'm guessing this is because I'm on Swift 5.7. I'm quite happy to upgrade to 5.8, would that be normal for the ecosystem or are a lot of people stuck on older versions of Swift?

heckj: Yeah, that's entirely the case. The options I tacked on to the end are swift 5.8 specific features. Since it's not yet released, and you're asserting a Swift 5.6 base for the swift side of this, I'll pull those out. We can always add it back in later if that's appealing. I can host a local copy if you want to see the results.

PR #3, opened 2023-03-07
updating preview script to use package structure and plugin

heckj:
uses docc-plugin to preview documentation


alexjg:
Running `preview-docs.sh` on my machine with these changes no longer shows the additional files in the documentation catalogue (the structure in `Extension.md` and the `AddressBookExample.md`), is there some additional incantation I need to make that appear?

heckj:
Let me catch up to the merges and double-check - I thought they were in there, but I might have missed something in the separate merges

alexjg:
No wait, I'm in the wrong here, I was looking at the wrong branch, my apologies. Instead I get this error:

```
Building for debugging...
Build complete! (0.15s)
Building for debugging...
[14/14] Compiling Automerge value.swift
Build complete! (0.93s)
error: 'Automerge' does not contain any documentable symbols or a DocC catalog and will not produce documentation
```

I think this may be a bug in swift 5.7?

(note, apologies for closing and reopening and adding a rogue comment, my keyboard appears to be possessed)

PR #7, opened 2023-03-08
doc: Adds documentation abstracts

heckj:
- also updated any references of `automerge` to `Automerge` in the narrative to reflect it being a proper noun in other Automerge documentation.
- removed a redundancy in the doc build script, and documented an upstream bug with swift-docc-tools and using a local XCFramework reference (as this project does)

### DIFF
--- a/Sources/Automerge/actor_id.swift
+++ b/Sources/Automerge/actor_id.swift
@@ -1,5 +1,6 @@
 import AutomergeUniffi
 
+/// A type that represents a collaboration instance for managing replication within an Automerge document.
 public struct ActorId: Equatable, Hashable {
     internal var bytes: [UInt8]
 }
--- a/Sources/Automerge/actor_id.swift
+++ b/Sources/Automerge/actor_id.swift
@@ -1,6 +1,6 @@
 import AutomergeUniffi
 
-/// A type that represents an identifier for a collaborator that makes changes within an Automerge document.
+/// The identifier for collaborators contributing to an Automerge document.
 ///
 /// Each separate instance of an Automerge document should have it's own, unique, `ActorId`.
 /// If you create your own `ActorId`, no concurrent changes should ever be made with the same `ActorId`.
--- a/Sources/Automerge/actor_id.swift
+++ b/Sources/Automerge/actor_id.swift
@@ -1,6 +1,9 @@
 import AutomergeUniffi
 
-/// A type that represents a collaboration instance for managing replication within an Automerge document.
+/// A type that represents an identifier for a collaborator that makes changes within an Automerge document.
+///
+/// Each separate instance of an Automerge document should have it's own, unique, `ActorId`.
+/// If you create your own `ActorId`, no concurrent changes should ever be made with the same `ActorId`.
 public struct ActorId: Equatable, Hashable {
     internal var bytes: [UInt8]
 }
--- a/Sources/Automerge/error.swift
+++ b/Sources/Automerge/error.swift
@@ -13,6 +13,9 @@ typealias FfiReceiveSyncError = AutomergeUniffi.ReceiveSyncError
 // binding generator and we would rather not couple our public API to the
 // specific binding generator tech we're using
 
+/// An general document error.
+///
+/// The error is specific to the Rust language binding infrastructure.
 public struct DocError: Error {
     let inner: FfiDocError
 
@@ -21,6 +24,9 @@ public struct DocError: Error {
     }
 }
 
+/// An error that indicates the synchronisation state could not be decoded.
+///
+/// The error is specific to the Rust language binding infrastructure.
 public struct DecodeSyncStateError: Error {
     let inner: FfiDecodeSyncStateError
 
@@ -29,6 +35,9 @@ public struct DecodeSyncStateError: Error {
     }
 }
 
+/// An error that indicates a problem loading the document.
+///
+/// The error is specific to the Rust language binding infrastructure.
 public struct LoadError: Error {
     let inner: FfiLoadError
 
@@ -37,6 +46,9 @@ public struct LoadError: Error {
     }
 }
 
+/// An error that indicates the received synchornisation could not be applied.
+///
+/// The error is specific to the Rust language binding infrastructure.
 public struct ReceiveSyncError: Error {
     let inner: FfiReceiveSyncError
 
--- a/Sources/Automerge/patch.swift
+++ b/Sources/Automerge/patch.swift
@@ -4,18 +4,18 @@ import enum AutomergeUniffi.PatchAction
 typealias FfiPatchAction = AutomergeUniffi.PatchAction
 typealias FfiPatch = AutomergeUniffi.Patch
 
-/// A collection of changes to an Automerge document.
+/// A collection of updates to the current state of an Automerge document.
 public struct Patch: Equatable {
-    /// The type change
+    /// The kind of update to apply.
     public let action: PatchAction
 
-    /// The path to the object identifier the action effects.
+    /// The path to the object identifier the update effects.
     public let path: [PathElement]
 
     /// Creates a new patch
     /// - Parameters:
-    ///   - action: The type of change
-    ///   - path: A collection of the paths to apply the change.
+    ///   - action: The kind of update to apply.
+    ///   - path: The path to the object identifier the update effects.
     public init(action: PatchAction, path: [PathElement]) {
         self.action = action
         self.path = path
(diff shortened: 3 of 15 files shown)


Sources/Automerge/actor_id.swift
```diff
@@ -1,5 +1,6 @@
 import AutomergeUniffi
 
+/// A type that represents a collaboration instance for managing replication within an Automerge document.
```
alexjg: This seems a little unclear to me, I'm not sure what "collaboration instance" refers to. Is this a known term in Swift? I've typically described the actor ID as representing a sequential thread of execution but il not sure that's a useful description either. The main thing is that no concurrent changes should be made by the same actor ID

Sources/Automerge/error.swift
```diff
@@ -13,6 +13,9 @@ typealias FfiReceiveSyncError = AutomergeUniffi.ReceiveSyncError
 // binding generator and we would rather not couple our public API to the
 // specific binding generator tech we're using
 
+/// An general document error.
+///
+/// The error is specific to the Rust language binding infrastructure.
```
alexjg, line 18: Is the purpose of the second paragraph here to explain why this isn't more descriptive (i.e. that it's not an enum with descriptive case names)?

Sources/Automerge/patch.swift
```diff
@@ -4,10 +4,18 @@ import enum AutomergeUniffi.PatchAction
 typealias FfiPatchAction = AutomergeUniffi.PatchAction
 typealias FfiPatch = AutomergeUniffi.Patch
 
+/// A collection of changes to an Automerge document.
```
alexjg: I think this is a bit confusing because we also refer to change hashes elsewhere, but the "change" in "change hash" is not the same as the change in a patch. (This is a symptom of some general muddiness around the design of the patch API I think)

A patch really represents a change in the observable state, rather than any change at all. It's intended to be used for applications to observe when something has happened that they might need to respond to.

Maybe a better name would be `Diff`?

In any case I don't think we should refer to them as "changes", just to avoid confusion.

Sources/Automerge/patch.swift
```diff
@@ -4,10 +4,18 @@ import enum AutomergeUniffi.PatchAction
...
+    /// The type change
     public let action: PatchAction
+
+    /// The path to the object identifier the action effects.
     public let path: [PathElement]
 
+    /// Creates a new patch
+    /// - Parameters:
+    ///   - action: The type of change
+    ///   - path: A collection of the paths to apply the change.
```
alexjg: The path isn't a collection of paths, it's a single path made up of `PathElement`s

Sources/Automerge/patch.swift
```diff
@@ -19,12 +27,19 @@ public struct Patch: Equatable {
     }
 }
 
+/// The type of change to apply along with the data to change.
```
alexjg: I think my comment above applies here as well and to the doc comments below. The patch is a notification that something has happened, not something a user creates to make something happen. On that basis I think the documentation should use the passive past tense rather than imperative present.

Sources/Automerge/patch.swift
```diff
@@ -19,12 +27,19 @@ public struct Patch: Equatable {
...
 public enum PatchAction: Equatable {
+    /// Put a new value at the property for the identified object.
     case Put(ObjId, Prop, Value)
+    /// Insert a collection of values at the index you provide for the identified object.
     case Insert(ObjId, UInt64, [Value])
+    /// Splices characters into and/or removes characters from the identified object at a given position within it.
```
alexjg, line 51: I suppose here we should probably mention that the index is a utf-8 code point index and not a grapheme cluster index

Sources/Automerge/path.swift
```diff
@@ -4,6 +4,7 @@ import enum AutomergeUniffi.Prop
 typealias FfiPathElem = AutomergeUniffi.PathElement
 typealias FfiProp = AutomergeUniffi.Prop
 
+/// The path to a specific object identifier within an Automerge document.
```
alexjg: I think rather than saying it's a path to an object identifier it might be better to say a "path to a property within an object"?

Sources/Automerge/scalar.swift
```diff
@@ -2,16 +2,27 @@ import enum AutomergeUniffi.ScalarValue
...
     case String(String)
+    /// An unsigned integer
     case Uint(UInt64)
+    /// A signed integer
     case Int(Int64)
+    /// A floating point number
     case F64(Double)
+    /// An integer counter
     case Counter(Int64)
+    /// A timestamp
```
alexjg: Might be worth specifying here that the value is milliseconds since the UNIX epoch

Sources/Automerge/scalar.swift
```diff
@@ -2,16 +2,27 @@ import enum AutomergeUniffi.ScalarValue
...
     case Timestamp(Int64)
+    /// A Boolean value
     case Boolean(Bool)
+    /// An unknown, raw scalar type.
```
alexjg, line 23: This type basically exists for forwards compatibility. Might be worth mentioning that it's not expected that users would construct this case

Sources/Automerge/actor_id.swift
heckj: Fair enough - I don't know what to call the idea of different places that are syncing together, and have been leaning on that term to avoid confusing it was "a person", "a machine", or even "a tab in a browser". I'll go re-read the general [Automerge docs](https://automerge.org/docs/hello/) and see if they have a term of art already established there and try to replicate that.

Sources/Automerge/error.swift
heckj: The way this'll show in DocC, the first line is the abstract that shows high level, and the second line will show in the reference page for this specific element as a second paragraph after the abstract. I've generally leaned on this location to provide additional detail for what the type is, does, or is used for beyond what's in the abstract - and repeating that these are all due to UniFFI/language bindings didn't seem relevant to the abstract itself.

I actually want to go back and re-read the uniffi generated bits to see how they're used to see if I can come up with better explanations of what they mean if they should appear, but thought this might be a reasonable placeholder to start with.

Sources/Automerge/patch.swift
heckj: Ah! Very good - I wasn't entirely sure and hadn't re-read all the code to know, so I was guessing based on the name alone. I'll rip out the "change" phrasing here entirely. I'll re-use the detail you just provided - and I think the name is a lot more relevant if you envision an analogue to "diff/patch" in the Git/source-control world.

Sources/Automerge/patch.swift
heckj: I'll update it - I was going just off the API signature that had that as a list of the type PathElement, and wasn't sure how that stitched together. That tidbit/confusion is also in my API questions list ;-) And I think you just cleared it up for me, although I might need a touch more detail about what a PathElement is, and how they're expected to be arranged and/or consumed. I'm guessing this is something a developer would more "want to react to" than construct themselves, but I wasn't sure.

Sources/Automerge/patch.swift
heckj: I really prefer to avoid any usage of passive voice, primarily because interpreting it can leave readers very muddy about what's taking the action. I'll definitely switch up the language, but I think we're far better off keeping it active voice, even if we use past tense to describe it.

Sources/Automerge/patch.swift
heckj: Definitely, I was guessing on the specifics of these internals (that's where I was reading the C API in the repo to get the details) - I'll make updates accordingly and make sure that's detailed here.

Sources/Automerge/scalar.swift
heckj: Good to know - I'll mark that down.

Sources/Automerge/scalar.swift
heckj: I'll make an explicit note of it being a reserved type.

heckj:
@alexjg I've given it that additional iteration after our chat earlier this week, so it should be good to review for full inclusion at this point.

Sources/Automerge/patch.swift
```diff
@@ -4,10 +4,29 @@ import enum AutomergeUniffi.PatchAction
...
     public let path: [PathElement]
 
+    /// Creates a new patch
+    /// - Parameters:
+    ///   - action: The kind of update to apply.
+    ///   - path: The path to the object identifier that the action effects.
+    ///
+    ///   The `path` does not identify the property on an object, or index in a sequence, that is updated, only the object that is effected.
+    ///   The `action` includes the type of change, and the value being updated, if relevant to the change.
     public init(action: PatchAction, path: [PathElement]) {
```
alexjg: I recall we discussed making this `internal` as this would never be constructed by the user. Can we do that?

heckj: Done in the follow-up commit, left the comments on the initializer anyway - but made it 'protected' so it's not accessible external to the module (i.e. developers using the package).

PR #8, opened 2023-03-09
2D benchmarks

heckj:
Uses swift-collections-benchmarks ([its documentation](https://github.com/apple/swift-collections-benchmark/blob/main/Documentation/01%20Getting%20Started.md)) to run benchmarks and generate images showing the "time to `x`" over the size of the collection. 

I stubbed in an additional test in the unit tests to make sure I understood how to "append" content to a document, which doesn't assert anything useful - my example text uses Emoji, and I immediately ran into the UTF-8 code points vs. Grapheme clusters bit on Text and lengths.

The benchmark names aren't substantively thought it - I was more focused on just getting them operational and see what the baseline numbers were. The numbers in the baseline results included in this PR are from my M1 MacBook Pro laptop - I'm certainly not asserting they should be authoritative, but they definitely were informative. Take a look yourself, but I'd say performance is *not* an issue for the benchmark cases that I implemented.

I've tried to include substantive examples and references to other benchmarks (used in Swift Std Lib) to share how to create additional benchmarks. It's not "hard", but neither is it well documented.

Before we merge this:
- What sort of naming convention do we want to use?
- What baseline benchmarks would we like to have in place? (or should we merge this _as is_ and add separate benchmarks in their own PRs?
- What host do we want to use for running the benchmarks? Right now, I think consistency is most important - and previous benchmarks have run from your own system, so do we want to stay with that?
- Would you like more explicit documentation around being able to run a comparison against the baseline interactively when/if we've updated any relevant underlying code? (either the wrappers, or more likely the underlying Automerge library)

I can also pretty quickly add a dead simple single dimensional Benchmark example (which uses a different library), if that's of interest. Since it's one-dimensional, it's reporting data into an HdrHistogram - which is nice to see long-tail percentiles - but also has some nice integration tooling to verify "in CI" against baselines to hold against regressions. The swift-collections-benchmark has no real provisions there, focused more on providing visuals to be able to compare.


alexjg:
This looks great to me and it's very encouraging that the benchmarks that are here suggest we haven't gotten anything egregiously wrong in the process of exposing things via FFI.

The only thing I would add is a note in the top level README under the "Building and developing" section just mentioning that there are benchmarks we use in the `CollectionBenchmarks` folder.

As to the questions about naming and additional benchmarks and structure. I think for now these names are fine and we should add more benchmarks as and when we need them. To me the important point seems to be to have _some_ benchmarks we can run on an ad-hoc basis to make sure we aren't regressing horribly. Once we've developed some more of the higher level tooling we can start looking at specific cases which we need to focus on performance-wise. I've just always found it's easier to do that if someone has done the hard work of setting up a benchmark harness in the first place :pray:

alexjg (approved):

PR #9, opened 2023-03-13
replace [UInt8] in public API with Data

heckj:
The API transformation is quite direct, and primarily ergonomic - tests pass, but the API has changed with this.

One thing I noticed (and updated) was example code in the article `AddressBookExample.md`, which could easily get missed. They did get simpler with this update, but I'm thinking it might be a good idea to explicitly encapsulate some of those bits into "Snippets" - which are a relatively new Swift feature, specifically for Documentation, that provide tiny executables that can be referenced in articles and reference docs. They have the notable benefit of being explicitly compiled, so if there's an API change that would break a snippet that included in Documentation, this would flag it.

It's not quite as advanced as Rust (or Python)'s concept of DocTests - it doesn't check the output right now by running anything, only that it compiles - but still might be useful in these contexts, and it was pretty much made with the idea of mixing in pieces from a bit of example code into narrative just like the `AddressBookExample.md` articles does currently.

### DIFF
--- a/CollectionBenchmarks/Package.swift
+++ b/CollectionBenchmarks/Package.swift
@@ -7,7 +7,7 @@ let package = Package(
     platforms: [.macOS(.v10_15)],
     dependencies: [
         .package(url: "https://github.com/apple/swift-collections-benchmark", from: "0.0.1"),
-        .package(path: "../")
+        .package(path: "../"),
     ],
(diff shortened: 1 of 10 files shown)


alexjg:
This snippets thing sounds great! I had a brief look for something like it when I put this together

PR #18, opened 2023-04-07
Enabling Catalyst build

munhitsu:
This is an update to enable Catalyst target.
The caveat is that it requires a very specific pinned version of Rust (nightly-2023-02-02).


heckj (approved):
I'm a little concerned about the macabi support requiring an older version of Rust and only in nightly, but it seems like the immediate value outweighs that. Worst case we can drop macabi (Catalyst) support down the road, but in the meantime this makes it easier to adopt.

PR #23, opened 2023-04-16
Threading protection & strict Sendable checking

heckj:
I explored providing multi-thread protection for the Automerge API interface looking at Actors and async/await in Swift 5, but at the moment there's no clear path to handle it through there, as Swift language actors _are_ re-entrant and intentionally don't stop that flow, but the API requires it. It could be possible with actors and relevant locking to handle the re-entrant issues, but a simpler, apple-platform common technique is to barrier the API calls behind a serialization queue using Grand Central Dispatch.

I enabled this for Document, and am following through on marking the relevant types `Sendable` for future constraints and type checking within Swift.
- Added a serial dispatch queue to Document, and wrapped all calls through its queue
- Added a serial dispatch queue to SyncState, and wrapped all calls through its queue
- Marked some additional types as explicitly Sendable - SyncState and Document using `unchecked @Sendable`, and ChangeHash which didn't need any extra protection as a pure struct.
- Converted the generic collections of ChangeHash to more explicit Set<ChangeHash> to support strict Sendable checking for future-proof interactions.


heckj:
@alexjg Question for you. I switched up the generic collection to Set<ChangeHash>, but i'm wondering if the ordering of that set has any meaning or value to API consumers - such as the order of changes?

If that order should be maintained, there's an explicitly ordered set at https://github.com/apple/swift-collections/blob/main/Documentation/OrderedSet.md that we could use that guarantees the set/uniqueness constraints but also preserves the ordering if there's value or meaning in that.

wdyt?

alexjg:
Does this have performance implications? From what I understand of GCD the main thing to worry about here would be if some other process was starving the GCD threadpool then we might end up with latency on the calls to `queue.sync`? I'm not really concerned about this, more curious about the model :slightly_smiling_face: . Oh also, does this change any of our benchmarks?

As to the order of change hashes, definitely not significant, a set is a good model.

heckj:
DispatchQueues group work over threads, and don't necessarily spin up new threads for each group, but you're correct in that if there was a "thread explosion" in the app using this library, the contention there could limit through by thread starvation - but that would be pretty independent of the DispatchQueue bits here. The PR creates two serial, private dispatch queues - one for each class (SyncState and Document) that needs protecting for multithreaded access, and the overhead is expected to be exceptionally light. The QoS I chose (user-interactive) also notably the chances of thread starvation, as theses closures will be scheduled on threads (main, or otherwise nearby high priority threads) above everything background.

I really wanted to make this work with the new Swift concurrency primitives and the Actor model, but there's no clear path to dealing with re-entrancy with Actors at the moment, so a serialization process of some form made the most sense for general protection.

I'm using this as an excuse to knock together a script to compare benchmarks against the baseline, so I'll have some specific, hard numbers for you shortly (I hope) that verifies my hypothesis and expectation that the performance impact of this change is  negligible.

heckj:
Well, it definitely slowed things done a bit - but not (IMO) brutally. I script that I just added (takes a while to run!) shows the delta that I captured in a screenshot here. The list inserts all remained well below 10usec, and was consistently slower - all the various rendered graphs show the same. I'll attach the diff.html here too, in case that's readable - it looks like the graphs are SVGs encoded inline in the HTML in this case.

Screenshot 2023-04-18 at 8 15 25 PM

(diff.html zip'd so that Github would "accept" it)
[diff.html.zip](https://github.com/automerge/automerge-swifter/files/11268173/diff.html.zip)

alexjg:
Excellent, nothing like a pretty graph to start the morning 🙂. This looks entirely acceptable to me, it's not surprising that there is some impact, but adding a few microseconds to each operation is not a problem IMO.

alexjg (approved):

Issue #42, opened 2023-07-06
heads() as a set looses ordering info of history


heckj:
While exploring a possible "view history" feature of an example app, I realized that uniffi heads() returns a list, but in converting it to a set - we loose all the ordering of that history to show a sequence of changes. 

I believe a better result is either a straight up Array<ChangeHash>, preserving the ordering from the underlying library - or potentially bringing on Swift Collections OrderedSet to preserve the ordering from the underlying library.

alexjg:
Hmm. I'm not sure this is correct. The heads returned from e.g. `Automerge::get_heads` are not in any causal order, they are ordered lexicographically. The heads of a document don't really have a causal order as they represent concurrent changes.

heckj:
Well damn - I was hoping the ordering returned from the Rust layer was something I could use. I'm trying to make a slider that allows scrolling through changes  - and in my initial tests with just sequencing through the change sets resulted in all sorts of craziness - I assumed because we ditched the ordering. 

Is that something that just isn't feasible? I want to be able to see "changes are ordered in this way" - to scroll back in history without knowing prior changeset collections. Should I bag that idea?

alexjg:
We do have a topologically sorted list of changes in the document internally and that is exposed a little bit in the public Rust API. E.g. https://docs.rs/automerge/latest/automerge/struct.Automerge.html#method.get_changes returns the changes since a given set of heads in a topological order (note that this might be a confusing thing to scrub through a an linearization of a very branchy history will be weird).

So, we can expose that method in swift. It's not a nicely designed API for doing this stuff, but it might be enough to get started with?

heckj:
I'll go  get to reading, but in the meantime I'll revert this change. I got excited, made an assumption, and ran over what the current API exposed, assuming too much.

heckj:
resolved by #44 - added function `changes()` in addition to `heads()` and clarified what `heads()` actually returns in documentation.

PR #44, opened 2023-07-07
Adding a Get changes for causally ordered updates to the document

heckj:
@alexjg I could use some Rust specific help - I've got a function stubbed, but with a likely obvious compilation error. I'm not very fluent with Rust internals, nor idiomatic rust - so help on both/either would be very appreciated.

I added a function on the wrapper `doc` called `changes()` with the idea of getting all changes by providing an empty vector to `get_changes()` in Automerge. 

I'm getting the changes Vec<&Change>, but I'm not clear on how to convert that into Vec<ChangeHash> and the relatively obvious path of hoping there was an `.into()` available that I could call doesn't seem to be correct.

I also wasn't sure about returning the raw vector vs. something like `Result<Vec<ChangeHash>, DocError>` - and where errors would populate up from within the sequence of code to be entirely sure if an error _could_ be thrown from within what I wrote.

related to #42

### DIFF
--- a/Tests/AutomergeTests/TestHistory.swift
+++ b/Tests/AutomergeTests/TestHistory.swift
@@ -1,0 +1,29 @@
+import Automerge
+import XCTest
+
+class HistoryTests: XCTestCase {
+    func testChangeCountsInHeads() {
+        let doc = Document()
+        try! doc.put(obj: ObjId.ROOT, key: "key", value: .String("one"))
+
+        var heads = doc.heads()
+        XCTAssertEqual(heads.count, 1)
+        heads = doc.heads()
+        try! doc.put(obj: ObjId.ROOT, key: "key", value: .String("two"))
+        XCTAssertEqual(heads.count, 2)
+    }
+
+    func testChangeCountsInHeadsAndChanges() {
+        let doc = Document()
+        try! doc.put(obj: ObjId.ROOT, key: "key", value: .String("one"))
+
+        var heads = doc.heads()
+        XCTAssertEqual(heads.count, 1)
+        try! doc.put(obj: ObjId.ROOT, key: "key", value: .String("two"))
+        heads = doc.heads()
+        XCTAssertEqual(heads.count, 2)
+
+        let changes = doc.changes()
+        XCTAssertEqual(changes.count, heads.count)
+    }
+}
--- a/Tests/AutomergeTests/TestHistory.swift
+++ b/Tests/AutomergeTests/TestHistory.swift
@@ -9,36 +9,41 @@ class HistoryTests: XCTestCase {
 
         try doc.put(obj: ObjId.ROOT, key: "key", value: .String("two"))
         XCTAssertEqual(doc.heads().count, 1)
-        
+
         let replicaDoc = doc.fork()
         XCTAssertEqual(doc.heads().count, 1)
         XCTAssertEqual(replicaDoc.heads().count, 1)
         XCTAssertEqual(doc.heads(), replicaDoc.heads())
-        
-        try replicaDoc.put(obj: ObjId.ROOT, key: "key", value: .String("three"))
+
+        try doc.put(obj: ObjId.ROOT, key: "newkey", value: .String("beta"))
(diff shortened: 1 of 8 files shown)


rust/src/doc.rs
```diff
@@ -514,6 +514,13 @@ impl Doc {
             .collect()
     }
 
+    pub fn changes(&self) -> Vec<ChangeHash> {
+        let empty_heads: Vec<am::ChangeHash> = Vec::new();
+        let mut doc = self.0.write().unwrap();
+        let changes = doc.get_changes(&empty_heads);
+        changes.into_iter().map(|h| h.hash().into()).collect()
```
alexjg, line 521: The compiler in my head says that this should be `changes.into_iter().cloned().collect()` because the type of `changes` is `Vec<&ChangeHash>` so you need to clone the `&ChangeHash` to get a `ChangeHash`. I've not actually tried to compile it though so I might be completely off base.

heckj: the `get_changes` returned a Change - not a ChangeHash, which is how I jumped into the above formulation. It does compile in Rust, and then I had clippy and rustfmt yell at me until I tuned the syntax to where it was happy.

alexjg: Oh yes, sleepy brain is sleepy.

alexjg:
Re. error handling. I don't think we need to do any such thing here because `get_changes` is infallible.

Tests/AutomergeTests/TestHistory.swift
```diff
@@ -0,0 +1,29 @@
...
+class HistoryTests: XCTestCase {
+    func testChangeCountsInHeads() {
+        let doc = Document()
+        try! doc.put(obj: ObjId.ROOT, key: "key", value: .String("one"))
+
+        var heads = doc.heads()
+        XCTAssertEqual(heads.count, 1)
+        heads = doc.heads()
+        try! doc.put(obj: ObjId.ROOT, key: "key", value: .String("two"))
+        XCTAssertEqual(heads.count, 2)
```
alexjg: This should still be one I think. There are multiple heads when there are concurrent changes, for example if you were to fork the document then make a change in the fork, then merge the fork back into the original document you would have two heads. Then if you were to make another change in the original document you would be back to one head (which has the previous two heads as it's dependencies).

heckj: That explains why my test is failing ;-)

Tests/AutomergeTests/TestHistory.swift
```diff
@@ -2,28 +2,43 @@ import Automerge
...
+        XCTAssertEqual(replicaDoc.heads().count, 1)
+        XCTAssertEqual(doc.heads(), replicaDoc.heads())
+        
+        try replicaDoc.put(obj: ObjId.ROOT, key: "key", value: .String("three"))
+        XCTAssertEqual(doc.heads().count, 1)
+        XCTAssertEqual(replicaDoc.heads().count, 1)
+        XCTAssertNotEqual(doc.heads(), replicaDoc.heads())
+        
+        try doc.merge(other: replicaDoc)
+        XCTAssertEqual(doc.heads().count, 1)
```
alexjg: The reason this is 1 rather than 2 is that there were no _concurrent_ changes to `doc`. In git terms, the merge of `replicaDoc` into `doc` is a fast forward merge. If, after the fork but before the merge, you were to do `doc.put(...)` we _should_ see 2 heads.

heckj: AHHH!  Took a while to sink in to my skull, but that makes significantly more sense. With changes() in place and seeing the outputs, I think it's doing what I'm expecting/wanting (showing all changes to the document).

heckj:
@alexjg I *think* I've got this into something of a functional configuration. At first I screwed up and was calling heads() instead of changes(), but the tests are showing serial sequences of changes. I haven't yet tried to fully end-to-end use those as a values to get a document at each of those values as "heads", but I suspect I can use that to make a sort of "scroll through history" for the overall document.

PR #67, opened 2023-08-04
WIP - Feedback work

heckj:


rust/src/patches.rs
```diff
@@ -34,14 +34,24 @@ pub enum PatchAction {
...
     },
     SpliceText {
         obj: ObjId,
         index: u64,
         value: String,
         marks: HashMap<String, Value>,
     },
+    JoinBlock {
+        index: u64,
+        // ? cursor
```
heckj, line 46: @alexjg I got far enough to get things compiling, but not far enough to know how to reasonably translate these new Patch types out into Swift (or even much of a sense how they might be used). Struggled with what Cursor represents independently from index (and how to present it), as well as the Patch embedded within UpdateBlock.

heckj:
Reading through the PR and the relevant paper, pieces I'm still missing to enable this through to swift include:
- a representation of Block (a struct, perhaps with a name and array of parents - at least based on JS examples). Block may be equivalent to "Mark", but I'm not sure
- a function splitBlock (params: ObjId, index (UInt64), and a block instance) - and returns an ObjId
- a function joinBlock (params: ObjId)

The relevant rust functions for `joinBlock` and `splitBlock` use the type `ExId`, which may be? synonymous with ObjId, or translatable in some way, but I'm not sure what they are and how they relate as yet.

javascript variant for reference:
```js
  splitBlock(obj: ObjID, index: number, block: MapObjType): ObjID;
  joinBlock(block: ObjID) : void;
```

heckj:
Stuck in a bit of a conundrum - or perhaps just misunderstanding. It looks like to interact with blocks, I need to be able to invoke join_block and split_block, and the parameters for split_block reference a struct (loosely: `{ name:String, parents: Vec<String> }`) - but when I went to create an Automerge-swift library representation of the same, I found it wasn't part of the externally available API for the Automerge crate. I did take a stab at making it externally visible (akin to what was done with Mark), but immediately the compiler set to complaining about leaking other internal types (specifically `OpId`) - so now I'm doubly uncertain if an external representation is the right way to do this.

I've been inferring that split_block() and join_block() are the key API elements being added from reading the tests (typescript/javascript), but lack a lot of the details to trace how data flows in the same fashion that we're doing with UniFFI and the Automerge crate.

PR #91, opened 2023-12-18
SplitText operation seems not using UTF-8 encoding

miguelangel-dev:
> Maybe I am not consuming the API as it is expected.

Insert a single text `👨‍👩‍👧` into a new Automerge.Document is expected to have a length equal to `👨‍👩‍👧.utf8.count`. But it is retrieving a non-expected length.

After doing some checks I can confirm the next:
- Swift.UTF-8 👨‍👩‍👧 = 18 bytes; received: 5
- Swift.UTF-8 👩‍👩‍👧‍👦 = 25 bytes; received: 7
- Swift.UTF-8 🕰️ = 7 bytes; received: 2

### DIFF
--- a/Tests/AutomergeTests/TestText.swift
+++ b/Tests/AutomergeTests/TestText.swift
@@ -79,4 +79,14 @@ class TextTestCase: XCTestCase {
         // flaky assertion, don't do it :: because length is in UTF-8 codepoints, not!!! grapheme clusters.
         // XCTAssertEqual(stringLength, 5)
     }
+
+    func testEmojiInsertUTF8Length() throws {
+        let emoji = "👨‍👩‍👧"
+        let doc = Automerge.Document()
+        let textId = try doc.putObject(obj: ObjId.ROOT, key: "text", ty: .Text)
+
+        try doc.spliceText(obj: textId, start: 0, delete: 0, value: emoji)
+        let stringLength = doc.length(obj: textId)
+        XCTAssertEqual(stringLength, UInt64(emoji.utf8.count))
+    }
 }


heckj:
./cc @alexjg

Tests/AutomergeTests/TestText.swift
```diff
@@ -79,4 +79,14 @@ class TextTestCase: XCTestCase {
...
     }
+
+    func testEmojiInsertUTF8Length() throws {
+        let emoji = "👨‍👩‍👧"
+        let doc = Automerge.Document()
+        let textId = try doc.putObject(obj: ObjId.ROOT, key: "text", ty: .Text)
+
+        try doc.spliceText(obj: textId, start: 0, delete: 0, value: emoji)
+        let stringLength = doc.length(obj: textId)
+        XCTAssertEqual(stringLength, UInt64(emoji.utf8.count))
```
alexjg, line 90: I believe `utf8.count` here is the number of UTF-8 code units, whereas the length automerge returns is the number of unicode scalar values. This emoji  is `U+1F468,U+200D,U+1F469,U+200D,U+1F467`, which as you can see is 5 unicode scalars but will obviously need more than that number of  code units as a code unit is an 8 bit integer.

alexjg: I think you should be able to get the right count by doing `emoji.unicodeScalars.count`

jamztang: I can confirm that using unicodeScalar length works, here's a test case:

```swift
    func testEmojiDeleteScalarLength7() throws {
        let string = "🇬🇧👨‍👨‍👧‍👦😀"
        let doc = Automerge.Document()
        let textId = try doc.putObject(obj: ObjId.ROOT, key: "text", ty: .Text)

        let start = UInt64("🇬🇧".unicodeScalars.count)
        XCTAssertEqual(start, 2)
        let delete = Int64("👨‍👨‍👧‍👦".unicodeScalars.count)
        XCTAssertEqual(delete, 7)
        let end = UInt64("😀".unicodeScalars.count)
        XCTAssertEqual(end, 1)

        try doc.spliceText(obj: textId, start: 0, delete: 0, value: string)
        try doc.spliceText(obj: textId, start: start, delete: delete) // delete "👨‍👨‍👧‍👦"

        let stringLength = doc.length(obj: textId)
        XCTAssertEqual(stringLength, start + end)

        let text = try doc.text(obj: textId)
        XCTAssertEqual(text, "🇬🇧😀")
    }
```

@alexjg do you mind elaborating why do automerge use unicodeScalars count instead of character count? The problem I see is that it's not very straight forward to understand, and would allow consumer to delete partial characters and results in invalid string for display.

image

Also the same thing will allow Mark to be applied to partial characters, which results in data ambiguity, for example, half of the character has a "bold":"true" mark, and another half do not.

jamztang: Regardless of the decision, we might wanted to update the documentation here https://github.com/automerge/automerge-swift/blob/main/Sources/Automerge/Document.swift#L581-L598C16

alexjg: Yeah we should definitely update the documentation.

I think what you are referring to when you say "character" is a grapheme cluster, is that correct? In principle I think this library could use grapheme cluster indices, but it would introduce implementation complexities. Primarily the following

1. Automerge doesn't store an array of bytes representing a text CRDT. Instead we have a b-tree of operations, one for each unicode scalar that was inserted as well as tombstones for deleted  characters. In order to be fast we maintain indices on each node of the tree which count the width of the whole subtree starting at that node. Implementing grapheme cluster indexing then requires modifying all of the index logic to track the grapheme cluster width of each node - which is complex because a grapheme cluster can cross b-tree node boundaries and in general is not a local computation over a single operation
2. Many platforms do not edit in terms of grapheme clusters. For example, if you type the 4-person family emoji from your example into a text box in Firefox and then press backspace the emoji will not be deleted, it will just turn into a 3-person family, then pressing backspace again will turn it into two separate emoji, i.e. actually increasing the grapheme cluster count. This is important because when you receive changes from other peers automerge will emit a patch which describes how to modify the current version of the string to match the new value. This patch logic would need to be modified to take account of grapheme clusters and for the same reasons as (1) would be complex to do whilst staying fast.

The decision to use unicode scalars then is an expedient one. I actually think the way Swift does string indexing is nice and would prefer to use it, but most platforms (and most importantly Javascript) do not index strings by grapheme clusters and so we have not put much work into supporting this way of addressing text.

heckj: @jamztang What would have made this easier for you to understand up front?

The text there now is:
```
/// Splice characters into the text object you specify.
    ///
    /// - Parameters:
    ///   - obj: The identifier of the text object to update.
    ///   - start: The index position, in UTF-8 code points, where the function begins inserting or deleting.
    ///   - delete: The number of UTF-8 code points to delete from the `start` index.
    ///   If negative, the function deletes characters preceding `start` index, rather than following it.
    ///   - values: The characters to insert after the `start` index.
    ///
    /// With `spliceText`, the `start` and `delete` parameters represent UTF-8
    /// code point indexes. Swift string indexes represent grapheme clusters, but Automerge works
    /// in terms of UTF-8 code points. This means if you receive indices from other parts
    /// of the application which are swift string indices you need to convert them.
```

How could we make it clearer that we're not operating on 'Swift character (grapheme cluster) counts", but on the raw unicode code points for these index positions?

jamztang: @alexjg Yes we're refering to grapheme clusters.

1. Understand that changing indices could be very complex internally. Just an idea tho: not sure how different in storing lists instead of text, But if you could treat text as `[Character]` array then perhaps we technically reuse the list mechanism. Also in case people don't wanted to use Mark and replace with list `[Attributes]` stores in a list, having grapheme clusters as indices will make this relatively transition easy.

2. Not sure if many platforms are using unicodeScalars indices, with my limited knowledge Quill and NSString are using utf16 indices.

FYI, this is what it takes to convert indices to unicodeScalars because our main TextModel are using grapheme units). Not too difficult but not the most convenient to consume.

```swift
class TextModel
    let doc: Automerge.Document
    var content: String {
        try! doc.text(obj: textId) ?? ""
    }

    func addAttribute(name: String, value: ScalarValue, in range: Range<String.Index>) throws {
        // convert to unicodeScalars count
        let start = content[..<range.lowerBound].unicodeScalars.count
        let end = content[..<range.upperBound].unicodeScalars.count
        try! doc.mark(
            obj: textId,
            start: UInt64(start),
            end: UInt64(end),
            expand: .none,
            name: name,
            value: value
        )
    }

    // building an `[Attributes]` array in terms of grapheme clusters
    func attributes() -> [RichTextAttributes] {
        guard let marks = try? doc.marks(obj: textId)
        else { return [] }
        var attributes: [RichTextAttributes] = Array(repeating: .init(), count: content.count)
        marks.forEach { mark in
            // convert mark.start and mark.end to grapheme indices
            let start = content.unicodeScalars.index(content.unicodeScalars.startIndex, offsetBy: Int(mark.start))
            let end = content.unicodeScalars.index(content.unicodeScalars.startIndex, offsetBy: Int(mark.end))
            let first = String(content.unicodeScalars[..<start])
            let string = String(content.unicodeScalars[start..<end])
            let range = Int(first.count)..<Int(string.count + first.count)

            let newAttributes = attributes[range].map { attributes in
                var attributes = attributes
                attributes.merge(mark.attributes)
                return attributes
            }
            attributes.replaceSubrange(range, with: newAttributes)
        }
        return attributes
    }
```

@heckj The main confusion with the text is referring to "UTF-8 code points". We needed to use `string.unicodeScalars` instead of `string.utf8` so perhaps updating that part into "UnicodeScalars code points" might be beneficial.

alexjg: Unfortunately returning a `[Character]` has basically the same problem as a `Character` is a grapheme cluster. The underlying data structure is effectively a sequence of unicode scalars, producing grapheme cluster indexes means grouping those scalars in some manner. We could efficiently produce a sequence of `Character` when reading a string, but using `Character` indices in patches and inserts is more complex. I will think harder about it though because as I said, I think grapheme cluster indexing is in general better.

Regarding the `Attribute` stuff. My current thinking is that the most convenient interface would actually be a diff based one. I.e. some kind of `Document.updateText(newText: AttributedString)` interface which then figures out the difference between the old attributes and the new attributes and generates the required change. This will require a bit of thought but I think that provided you call `updateText` with every keystroke or formatting change this would basically be equivalent and fits much more naturally with the APIs that UIKit and SwiftUI expose.

heckj:
Thank you for this example setup - I've incorporated it into #99, and with it found a number of issues elsewhere in the higher level apis that were using the incorrect assumptions about UTF-8 code points, which are also resolved with #99. I'm closing this PR, with the relevant code sample illustrating what's working in #99 - and with multiple verifications there.

miguelangel-dev:
@heckj Thanks for all the work around it, and the new version - it is really helpful

PR #92, opened 2024-01-10
Support WebAssembly

kateinoigakukun:
This PR port this project to WebAssembly by following changes:

- Use SwiftPM's Clang module integration instead of XCFramework on non-darwin platforms
- Make `Combine`, `SwiftUI`, `os.log`, and `libdispatch` optional dependencies since they are not available on WebAssembly and non-darwin platforms.

## Demo

See https://github.com/kateinoigakukun/automerge-swift-wasm

Users need to build `libuniffi_automerge.a` by themselves by `cargo build --manifest-path rust/Cargo.toml --target wasm32-wasi --release` and link it by adding linker option.


alexjg:
This is very exciting! If I understand correctly the `libuniffi_automerge.a` is an archive containing WASM objects, which means it should be platform independent. Can we make things easier for people by compiling it and hosting it as part of a Github release so that users can just grab it from the release URL?

kateinoigakukun:
@alexjg Exactly yes! Users will be able to skip cargo build if we will host pre-compiled static archives.
Unfortunately, it's not as easy as XCFramework binaryTarget, but much better than nothing 👍

heckj:
@kateinoigakukun Thank you - finally had a chance to really sit down with the PR this morning. I've updated it slightly to preserve the generated `automergeFFI.h` from UniFFI rather than using a completely static one - no idea if/when it will update, but I know I'd otherwise miss it and be deviled in debugging that missing link if I didn't.

I want to specifically add some CI work to compile this in WASM during builds to make sure we maintain the canImport() and related bits correctly for a WASM distribution down the road - do you have any suggestions (or examples to easily borrow from and replicate) to set something like that up with GithubActions? I'm guessing a Linux container build step might be easiest, but I just wasn't sure what the current "state of the art" was in setting up a CI with a swift-wasm compilation toolchain on it.

kateinoigakukun:
@heckj Thanks! I added wasm workflow here.

kateinoigakukun:
🎉

Issue #104, opened 2024-01-21
Accessing history timestamp


lightsprint09:
As far as I know, every change encodes a wall clock time stamp. Is this already exposed?

heckj:
That's not exposed from the core library (as of version 0.5.7)

alexjg:
It is possible to set the timestamp on a `Transaction` in the core library, but we are not using transactions here, we are using `AutoCommit`. I think what we should do is to add a configuration option to `AutoCommit` to allow you to annotate changes with a timestamp etc. (there are other metadata features we will want soon I think). Once we have that API we can expose the same API in Swift.

alexjg:
I've added an issue in the `automerge` repo here: https://github.com/automerge/automerge/issues/847

lightsprint09:
Still no progress on this?

Not sure what this[ line of code](https://github.com/automerge/automerge-swift/blob/57cdc2bcc9b90127f502ac829dfbec10285a76c0/Sources/Automerge/Change.swift#L25) does at the moment. Maybe we should remove the timestamp API for now

alexjg:
Reading through this issue I realise that I may have misunderstood the original problem. Is the problem that you want to be able to _read_ the timestamp on changes in the document or that you want to _set_ the timestamp?

lightsprint09:
I just want to read it. I guess it sets the value to something like `now()`?

alexjg:
I believe the line you are referring to does return the timestamp on the change if it was set (I have not checked this, just from reading the code and my hazy memories). The timestamps are not actually set in this library unless you use `Doc.commitWith` to set the timestamp though.

PR #129, opened 2024-03-13
Added commitWith() method to specify message/timestamp and exposing changes via changeByHash()

bgomberg:
We currently (with the older version of the swift bindings) use the `timestamp` field of the changes in order to determine when a document was last modified for display to the user in our application. I was directed to #104 where this feature has also been discussed. While changes to the underlying `AutoCommit` type in the core automerge library would likely be better long-term, this PR aims to provide a more targeted solution for this swift library. In the normal usage of this library with an `AutomergeEncoder`, the underlying transaction is made at the point that save() is called. Therefore, we can make this transaction more explicit and pass through commit options (`message` / `timestamp`).

I also needed a way to retrieve these commit options later, so added a new `changeByHash(hash:)` method which grabs a full `Change` object from the underlying automerge document for a given hash. This way, we can easily get the timestamp for the latest changes by calling this method on each `head` of the document.

I added a new test which showcases both of these new APIs.

As an aside, when I ran the `build-xcframework.sh` script locally, I ended up with a ton of other local changes to the generated code (mostly style changes). I removed those from the PR but would be great to know if there's a way to make this less noisy moving forwards.

### DIFF
--- a/AutomergeUniffi/automerge.swift
+++ b/AutomergeUniffi/automerge.swift
@@ -506,6 +506,8 @@ public protocol DocProtocol: AnyObject {
 
     func save() -> [UInt8]
 
+    func saveWithOptions(msg: String, time: Int64)  -> [UInt8]
+
     func setActor(actor: ActorId)
 
     func splice(obj: ObjId, start: UInt64, delete: Int64, values: [ScalarValue]) throws
@@ -521,6 +523,7 @@ public protocol DocProtocol: AnyObject {
     func values(obj: ObjId) throws -> [Value]
 
     func valuesAt(obj: ObjId, heads: [ChangeHash]) throws -> [Value]
+
 }
 
 public class Doc:
@@ -564,6 +567,7 @@ public class Doc:
             )
         })
     }
+
 
     public func actorId() -> ActorId {
         try! FfiConverterTypeActorId.lift(
@@ -1227,6 +1231,19 @@ public class Doc:
                     uniffi_uniffi_automerge_fn_method_doc_save(
                         self.uniffiClonePointer(),
                         $0
+                    )
+                }
+        )
+    }
+
+    public func saveWithOptions(msg: String, time: Int64)  -> [UInt8] {
+        try! FfiConverterSequenceUInt8.lift(
+            try!
+                rustCall {
+                    uniffi_uniffi_automerge_fn_method_doc_save_with_options(
+                        self.uniffiClonePointer(),
+                        FfiConverterString.lower(msg),
+                        FfiConverterInt64.lower(time),$0
                     )
                 }
         )
@@ -2968,6 +2985,9 @@ private var initializationResult: InitializationResult {
         return InitializationResult.apiChecksumMismatch
     }
     if uniffi_uniffi_automerge_checksum_method_doc_save() != 20308 {
+        return InitializationResult.apiChecksumMismatch
+    }
+    if uniffi_uniffi_automerge_checksum_method_doc_save_with_options() != 11279 {
         return InitializationResult.apiChecksumMismatch
     }
     if uniffi_uniffi_automerge_checksum_method_doc_set_actor() != 64337 {
--- a/AutomergeUniffi/automerge.swift
+++ b/AutomergeUniffi/automerge.swift
@@ -420,6 +420,8 @@ public protocol DocProtocol: AnyObject {
 
     func changeByHash(hash: ChangeHash) -> Change?
 
+    func commitWith(msg: String?, time: Int64)
+
     func cursor(obj: ObjId, position: UInt64) throws -> Cursor
 
     func cursorAt(obj: ObjId, position: UInt64, heads: [ChangeHash]) throws -> Cursor
@@ -507,8 +509,6 @@ public protocol DocProtocol: AnyObject {
     func receiveSyncMessageWithPatches(state: SyncState, msg: [UInt8]) throws -> [Patch]
 
     func save() -> [UInt8]
-
-    func saveWithOptions(msg: String, time: Int64) -> [UInt8]
 
     func setActor(actor: ActorId)
 
@@ -1238,6 +1238,18 @@ public class Doc:
             }
         )
     }
+    public func commitWith(msg: String?, time: Int64) {
+        try!
+            rustCall {
+                uniffi_uniffi_automerge_fn_method_doc_commit_with(
+                    self.uniffiClonePointer(),
+
+                    FfiConverterOptionString.lower(msg),
+                    FfiConverterInt64.lower(time),
+                    $0
+                )
+        }
+    }
 
     public func save() -> [UInt8] {
         try! FfiConverterSequenceUInt8.lift(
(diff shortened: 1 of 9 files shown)


AutomergeUniffi/automerge.swift
```diff
@@ -1495,6 +1527,96 @@ public func FfiConverterTypeSyncState_lower(_ value: SyncState) -> UnsafeMutable
     FfiConverterTypeSyncState.lower(value)
 }
 
+public struct Change {
+    public var actorId: ActorId
+    public var message: String?
+    public var deps: [ChangeHash]
+    public var timestamp: Int64
+    public var bytes: [UInt8]
```
heckj, line 1534: I'd prefer that we expose this as Data (for these purposes, it's as simple as Data([uint8]) wrapper before dropping it out - which keeps it consistent with the external representation of the other parts of the API.

Sources/Automerge/Document.swift
```diff
@@ -922,6 +940,18 @@ public final class Document: @unchecked Sendable {
         }
     }
 
+    /// Returns an list of changes that represent the causal sequence of changes to the document.
```
```suggestion
    /// Returns the contents of the change associated with the change hash you provide.
```

Sources/Automerge/Document.swift
```diff
@@ -922,6 +940,18 @@ public final class Document: @unchecked Sendable {
...
+    ///
+    /// - Returns: A``Change`` object for the given hash.
```
```suggestion
```

Sources/Automerge/Document.swift
```diff
@@ -783,6 +783,24 @@ public final class Document: @unchecked Sendable {
         }
     }
 
+    /// Encode the Automerge document in a compressed binary format.
```
heckj, line 785:
```suggestion
    /// Encode the Automerge document in a compressed binary format with additional metadata.
```

Sources/Automerge/Document.swift
```diff
@@ -783,6 +783,24 @@ public final class Document: @unchecked Sendable {
...
+    ///
+    /// - Parameters:
+    ///   - message: A message to attach to the auto-committed change (if any).
```
```suggestion
    ///   - message: An optional message to attach to the compacted change.
```

Sources/Automerge/Document.swift
```diff
@@ -783,6 +783,24 @@ public final class Document: @unchecked Sendable {
...
+    ///   - timestamp: The timestamp to attach to the auto-committed change (if any).
```
```suggestion
    ///   - timestamp: An optional timestamp to attach to the auto-committed change.
```

Sources/Automerge/Document.swift
```diff
@@ -783,6 +783,24 @@ public final class Document: @unchecked Sendable {
...
+    /// - Returns: The data that represents all the changes within this document.
+    ///
+    /// The `saveWithOptions` function also compacts the memory footprint of an Automerge document and increments the
+    /// result of ``heads()``, which indicates a specific point in time for the history of the document.
+    public func saveWithOptions(message: String, timestamp: Date) -> Data {
```
```suggestion
    public func saveWithOptions(message: String?, timestamp: Date?) -> Data {
```

Sources/Automerge/Document.swift
```diff
@@ -783,6 +783,24 @@ public final class Document: @unchecked Sendable {
...
+        sync {
+            self.doc.wrapErrors {
+                sendObjectWillChange()
+                return Data($0.saveWithOptions(msg: message, time: Int64(timestamp.timeIntervalSince1970)))
```
heckj: Given the description of the API, I think it makes sense to allow one or both of these additional metadata pieces to be null - so we should accommodate that as well. Convert an empty into a null string for the initializer, or just pass the nil if it's set as such, and a missing timestamp into - perhaps? - the current timestamp using something like a new Date() initializer

heckj (changes requested):
Conceptually, looks good to me, but I'd like @alexjg to chime in on the Rust code and any implications for exposing the Change contents that I might be missing. 

Was a little surprised you replicated/shadowed Change into Automerge-swift rather than using extensions to add pieces to the underlying one, but no qualms with how you did it - it's explicit and clear on what's happening there.

For the optional metadata save component of this, I would like to make either, or other, of those parameters actually optional and handle the missing pieces inside that method appopriately. Ideally, there'd be tests illustrating and working passing in an optional message, an optional timestamp, or both (which seems pathologically silly - but possible) - and what that means (for example - does calling this method with a nil date mean "please save the current timestamp", or is it functionally the same as calling the non-optional-enabled `save()` method?)

I added some suggested updates to documentation changes on the save method - there's changes that'll be needed in overall curation and other documentation locations, but I can handle integrating those post-merge without issue.

AutomergeUniffi/automerge.swift
```diff
@@ -1495,6 +1527,96 @@ public func FfiConverterTypeSyncState_lower(_ value: SyncState) -> UnsafeMutable
...
+    public var hash: ChangeHash
+
+    // Default memberwise initializers are never public by default, so we
+    // declare one manually.
+    public init(
```
alexjg, line 1539: Is there a reason to have this public? It feels like it should only ever be used internally.

rust/src/doc.rs
```diff
@@ -500,6 +500,14 @@ impl Doc {
         doc.save()
     }
 
+    pub fn save_with_options(&self, message: String, time: i64) -> Vec<u8> {
```
alexjg: What's the reasoning behind combining `save` and `commit`? It feels to me like a nicer API would be to expose a `commit_with` method which just calls `doc.commit_with(..)`.

AutomergeUniffi/automerge.swift
heckj: (I had the same question when I read through, but realized later that he set up a replica struct in Automerge with the same name and initializes it from this type, rather than passing through the type from AutomergeUniffi and adding extensions onto it to provide the additional equality and hashable conformances)

alexjg: Ah gotcha.

Sources/Automerge/Document.swift
bgomberg: Will make `message` optional since that can be null on the underlying Rust type. I think giving the timestamp a default `Date().timeIntervalSince1970` value is the most convenient.

AutomergeUniffi/automerge.swift
bgomberg: Good idea - will add this to `Automerge.Change`

bgomberg:
Thanks for the feedback @heckj @alexjg.

I separated out `commitWith()` as that does seem cleaner than overloading `save()` like I had before. One thing that's a bit funky, but feels like the behavior I would want given the underlying Rust implementation, is that calling `save()` results in a `nil` message and `0` timestamp, whereas calling `commitWith()` results in a `nil` message and `Date().timeIntervalSince1970` timestamp. I do like having the Swift API populate the date for me, and would personally vote for having the Rust code do the same by default (and thus move the `save()` behavior to match `commitWith()`, but could understand not wanting to add that dependency. Regardless, I updated the tests to explicitly exercise and demonstrate the differences.

Sources/Automerge/Document.swift
```diff
@@ -768,6 +768,23 @@ public final class Document: @unchecked Sendable {
         }
     }
 
+    /// Commit the auto-generated transaction with options.
+    ///
+    /// - Parameters:
+    ///   - message: An optional message to attach to the auto-committed change (if any).
+    ///   - timestamp: An optional timestamp to attach to the auto-committed change (if any).
+    ///
+    /// The `commitWith` function also compacts the memory footprint of an Automerge document and increments the
```
alexjg: I think this doc comment is no longer accurate. `commitWith` doesn't compact the memory in any way. It just commits any uncommitted changes. Furthermore, if there are no outstanding changes (i.e. you haven't actually written anything to the doc since the last time you committed, or saved, or received a sync message etc.) then the heads of the document will not change.

alexjg:
Re. the `nil` message and `0` timestamp. If you call `commitWith` followed by `save` then the `save` shouldn't actually create a new commit. The reason is that `save` (and all the other methods which trigger a commit) don't do anything unless there are actually outstanding changes, so if you've already saved everything then nothing should happen.

heckj (approved):

PR #132, opened 2024-03-15
Added public initializer to ActorId and tests to validate setting `actor` property

bgomberg:
Regarding the use-case for this, we set the actorId by concatenating the user ID with a globally-unique device ID. We leverage this heavily in our internal tooling to browse document history for debugging various issues. Here's an example where can see who made each change by looking up the user ID. This is very helpful both for development to figure out issues with our sync protocol, as well as customer support in figuring out what made what change to a document.

Screenshot_2024-03-15_at_4 46 26_PM

### DIFF
--- a/Sources/Automerge/ActorId.swift
+++ b/Sources/Automerge/ActorId.swift
@@ -8,6 +8,17 @@ import Foundation
 public struct ActorId: Equatable, Hashable, Sendable {
     var data: Data
 
+    // Creates a new, random actor.
+    public init() {
+        self.init(uuid: UUID())
+    }
+
+    // Creates an actor from a UUID.
+    public init(uuid: UUID) {
+        data = withUnsafeBytes(of: uuid.uuid, { Data($0) })
+    }
+
+    // Creates an actor from data.
     public init(data: Data) {
         self.data = data
     }
--- a/Sources/Automerge/ActorId.swift
+++ b/Sources/Automerge/ActorId.swift
@@ -6,6 +6,10 @@ import AutomergeUniffi
 /// If you create your own `ActorId`, no concurrent changes should ever be made with the same `ActorId`.
 public struct ActorId: Equatable, Hashable, Sendable {
     var bytes: [UInt8]
+
+    public init(bytes: [UInt8]) {
+        self.bytes = bytes
+    }
 }
 
 extension ActorId: CustomStringConvertible {
--- a/Sources/Automerge/ActorId.swift
+++ b/Sources/Automerge/ActorId.swift
@@ -8,6 +8,10 @@ import Foundation
 public struct ActorId: Equatable, Hashable, Sendable {
     var data: Data
 
+    init(ffi: AutomergeUniffi.ActorId) {
+        data = Data(ffi)
+    }
+
     // Creates a new, random actor.
     public init() {
         self.init(uuid: UUID())
@@ -19,7 +23,10 @@ public struct ActorId: Equatable, Hashable, Sendable {
     }
 
     // Creates an actor from data.
-    public init(data: Data) {
+    public init?(data: Data) {
+        guard data.count <= 128 else {
+            return nil
+        }
         self.data = data
     }
 }
--- a/Sources/Automerge/ActorId.swift
+++ b/Sources/Automerge/ActorId.swift
@@ -1,19 +1,20 @@
 import AutomergeUniffi
+import Foundation
 
 /// The identifier for collaborators contributing to an Automerge document.
 ///
 /// Each separate instance of an Automerge document should have it's own, unique, `ActorId`.
 /// If you create your own `ActorId`, no concurrent changes should ever be made with the same `ActorId`.
 public struct ActorId: Equatable, Hashable, Sendable {
-    var bytes: [UInt8]
+    var data: Data
 
-    public init(bytes: [UInt8]) {
-        self.bytes = bytes
+    public init(data: Data) {
+        self.data = data
     }
 }
 
 extension ActorId: CustomStringConvertible {
     public var description: String {
-        bytes.map { Swift.String(format: "%02hhx", $0) }.joined().uppercased()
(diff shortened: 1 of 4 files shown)


heckj:
Thanks @bgomberg  - the [ActorId struct](https://github.com/automerge/automerge/blob/main/rust/automerge/src/types.rs#L39) in the core library is an explicit 16 bytes. Swift doesn't have _quite_ as easy a path to assert that a type is a specific length of bytes, and I'm not sure what happens if you attempt to initialize it with something significantly larger, and if it supports inter-operation with documents with different actorId sizes. It may be that with the current implementation, we're constrained to only 16 bytes, and that if you do implement it, you're responsible for verifying you're not making concurrent updates with the same actorId.

If we're going to expose a public initializer, then I think we'd want at least constrain it to 16 bytes. I'm doing the same in DocumentId with the Automerge-repo work ([implementation external from here](https://github.com/automerge/MeetingNotes/blob/automerge-repo-package/Packages/automerge-repo/Sources/AutomergeRepo/DocumentId.swift)) - the gist of which is a default initializer that loads from a UUID, and any other initializers that verify the content is 16 bytes or kicks out an error.

Sources/Automerge/ActorId.swift
```diff
@@ -1,15 +1,20 @@
...
 
 /// The identifier for collaborators contributing to an Automerge document.
 ///
 /// Each separate instance of an Automerge document should have it's own, unique, `ActorId`.
 /// If you create your own `ActorId`, no concurrent changes should ever be made with the same `ActorId`.
 public struct ActorId: Equatable, Hashable, Sendable {
-    var bytes: [UInt8]
+    var data: Data
+
+    public init(data: Data) {
```
heckj: If we're going to have an initializer, I'd prefer to have a default initializer that doesn't require any input, to minimize the use of it - but allows you to pass in a 16-byte data buffer to support your use case and needs. I think it also needs to be a failable initializer that throws an error if it's other than 16 bytes.

Since explicit length arrays are kind of a PITA in swift, I'd recommend something like:


```swift
public init?(data: Data? = nil) {
// ... check for explicitly 16 bytes, and throw an error if its not
}
```

I'm asserting this based on the idea that the Automerge core _doesn 't_ support anything other than 16 byte actorIds, but I'm not 100% sure on that point. @alexjg would be best to weigh in there. I think that's what I read on the Rust source, but I'm not super confident that I didn't miss some alternate trait that allows for variable lengths to be used somewhere.

heckj (changes requested):

Sources/Automerge/ActorId.swift
bgomberg: We're currently using 32-byte actorIds (since they are a concatenation of 2 UUIDs), so I would hope the newer versions of the Rust code isn't enforcing 16 bytes for backwards compatibility at least. I did a quick look through the Rust code and do see that the actor ID is specified as 16 bytes, but also see a case where it's being set to 20, so doesn't seem like a hard limit at least.

heckj: Let's definitely start by adding a test to make sure it round trips from initializer to returning a full 32 bytes for your example setup, and we can pin down a "reasonable max size" after. If not, it'll likely mean core work in the Rust library to accommodate the size differences before we can finalize what you need here in the Swift overlay.

If you don't mind, check the PR to allow for updates, and I'll help add documentation and such to this PR, rather than trying to slap it in after the fact - as well as some of the relevant tests that I'd like to see, at least if you're amenable.

bgomberg: > Let's definitely start by adding a test to make sure it round trips from initializer to returning a full 32 bytes for your example setup, and we can pin down a "reasonable max size" after. If not, it'll likely mean core work in the Rust library to accommodate the size differences before we can finalize what you need here in the Swift overlay.

Updated with a test.

> If you don't mind, check the PR to allow for updates, and I'll help add documentation and such to this PR, rather than trying to slap it in after the fact - as well as some of the relevant tests that I'd like to see, at least if you're amenable.

Is this something I need to set on the PR? Or you just mean to merge suggested changes?

heckj: This detail: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

I'm returning to Seattle tomorrow, so I won't add anything straight-away, but I thought I might add docs and tests as commits to your pull request branch, if you allow it.

alexjg: The `ActorId` struct is not hardcoded to a specific lenght, however if the actor ID is 16 bytes or less then it can be stack allocated in many cases which has a small performance impact.

bgomberg: @heckj I'm not seeing that option, and as far as I can tell, I think that option isn't available here since I created the fork as part of my organization and not my own personal user.

heckj: no worries @bgomberg - I'll work on putting suggestions into the PR review process itself. It's hard to ask for added lines, etc - but we'll work it out, and I can do additional docs to what I'd like to see after the fact if needed.

Sources/Automerge/ActorId.swift
```diff
@@ -1,15 +1,31 @@
...
+    public init() {
+        self.init(uuid: UUID())
+    }
+
+    // Creates an actor from a UUID.
+    public init(uuid: UUID) {
+        data = withUnsafeBytes(of: uuid.uuid, { Data($0) })
+    }
+
+    // Creates an actor from data.
```
heckj, line 25: Pick a max size - I'm not stressed about how max, but this _should_ have a limit - and add a check, failing the initializer if it's over a max size. You've mentioned you need 32 bytes for your use case. So as a suggestion, how about 128 bytes as a max?

heckj (approved):
Thanks @bgomberg !

PR #136, opened 2024-03-31
Adds support to encode & decode Foundation.URL

lightsprint09:


heckj (approved):
Hey Lucas! Looks good - thanks for the submission.

I admit I didn't check explicitly, but I thought URL was already codable and that it's default encoding would work to do what you're after. Is it not?

lightsprint09:
No worries, would have guessed the same. I am currently migrating from the old automerge to the new one and discovered it did not work as expected.

heckj:
yeah, just double checked and URL is definitely Codable - I'm guessing this was because you wanted a specific in-automerge-schema representation for it? That plain string?

When I coded it (prior to this PR) and dumped the schema, I got:
```
      "url" : {
        {
        "relative" :String(https://automerge.org/)
        }
      }
```

lightsprint09:
Right, I would expect. This is how JSONDecoder/Encoder handle an URL

```
"url" : String(https://automerge.org/)
```

PR #137, opened 2024-03-31
Removes force downcast in favor of condidional cast in switch

lightsprint09:
While implementing URL support I found an elegant way to get rid of force casting in EncodingConatiners. Should not be slower, maybe even faster.


heckj:
Looks good - drops to a single comparison. Didn't benchmark it, but I expect it would be at least slightly faster.

PR #140, opened 2024-03-31
Improve Semantic of ScalarValueRepresentable

lightsprint09:
As the name describes ScalarValueRepresentable is all about Scalars, so I removed the value stuff.


heckj (approved):
Makes sense - I should have pulled that out some time ago (it was from earlier experiments to make something akin to what's now available as the Codable conformances)

Issue #139, opened 2024-03-31
Timestamp


lightsprint09:
The old automerge-swift stored Timestamp in milliseconds (not sure why), resulting in wrong resulting Dates when importing it with the new automerge-swift. Any idea we can mitigate this?

heckj:
It would require changes to the Automerge core - I'm uncertain on why it changed, exactly. I considered a transformation that preserved a Double (Date's internal TimeInterval type) from 1970 epoch, but in some ways that's even more oblique, and clearly doesn't map to the internal Automerge representation which has less granular resolution, so it didn't seem to offer any cross-platform benefit.

@alexjg - any insight into the change in internal Automerge representation of Value.Timestamp?

lightsprint09:
I guess the old auto merge implemented it wrong

alexjg:
I think this is actually a case where we got it wrong in the Swift implementation. The rest of the automerge implementations use _milliseconds_ since the epoch.

alexjg:
Ideally we would update the swift implementation (which I think just corresponds to changing the documentation?).

lightsprint09:
So the new implementation is missing this [piece of code](https://github.com/automerge/automerge-swift-archived/blob/4e57b78ea932c4abdf85002c0d2cde99c59614d7/Source/applyPatch.swift#L326)?

PR #141, opened 2024-04-01
Use millisecond semantic to match automerge-core

lightsprint09:
This changes timestamp to use milliseconds as documented [here](https://automerge.org/automerge-binary-format-spec/#_value) and discussed in #139 . Date is now used to represent a timestamp, to prevent mistakes when encoding/decoding. Converting Date to FFI representation is now in a single file

### DIFF
--- a/Sources/Automerge/ScalarValue.swift
+++ b/Sources/Automerge/ScalarValue.swift
@@ -17,7 +17,7 @@ public enum ScalarValue: Equatable, Hashable, Sendable {
     case F64(Double)
     /// An integer counter.
     case Counter(Int64)
-    /// A 64-bit signed integer representing milliseconds since the unix epoch
+    /// A 64-bit signed integer that represents the number of milliseconds since the UNIX epoch.
     case Timestamp(Date)
     /// A Boolean value.
     case Boolean(Bool)
--- a/Sources/Automerge/ScalarValue.swift
+++ b/Sources/Automerge/ScalarValue.swift
@@ -17,8 +17,8 @@ public enum ScalarValue: Equatable, Hashable, Sendable {
     case F64(Double)
     /// An integer counter.
     case Counter(Int64)
-    /// A timestamp represented by the number of seconds since UNIX epoch (Jan 1st, 1970, 00:00 UTC).
-    case Timestamp(Int64)
+    /// A 64-bit signed integer representing milliseconds since the unix epoch
+    case Timestamp(Date)
     /// A Boolean value.
     case Boolean(Bool)
     /// An unknown, raw scalar type.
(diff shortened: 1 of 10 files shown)


ept:
Hold on a minute, is it actually correct that automerge-core uses milliseconds resolution? I thought it was using seconds.

ept:
Oh sorry, we're talking about timestamps that appear in the document data. Yes, for those we use milliseconds. I thought we were talking about the change timestamp, which I believe uses integer seconds.

Sources/Automerge/ScalarValue.swift
```diff
@@ -17,8 +17,8 @@ public enum ScalarValue: Equatable, Hashable, Sendable {
     case F64(Double)
     /// An integer counter.
     case Counter(Int64)
-    /// A timestamp represented by the number of seconds since UNIX epoch (Jan 1st, 1970, 00:00 UTC).
-    case Timestamp(Int64)
+    /// A 64-bit signed integer representing milliseconds since the unix epoch
```
heckj:
```suggestion
    /// A 64-bit signed integer that represents the number of milliseconds since the UNIX epoch.
```

heckj (changes requested):
Thanks Lucas

Is there any place that we have a unified document with all the various data representations that we could use to verify the assertion that it's in milliseconds with Javascript and/or Rust? I'd really like to get _something_ that's concrete to verify the assertions of representation like this, especially since it was a mistake I made early on and jsut never caught, since I wasn't using this cross-platform - although that's the goal here.

heckj:
Note for release: we'll want to make sure to broadcast loudly that this changed - if anyone already has Date representations in their Automerge docs created, this will cause all of their Date values to be "broken", since we were representing them differently.

alexjg:
I also missed this so I think having some kind of concrete interop test would be a good idea, though I'm not sure exactly what that would look like.

alexjg:
In terms of written documentation the [value section](https://automerge.org/automerge-binary-format-spec/#_value) of the binary format docs exists but is quite dense.

heckj:
@alexjg I was thinking exactly the same thing just a few minutes earlier. Especially for the various language implementations, I think it would be a great idea. Maybe even just a document that has one of each of the .scalarValue data blobs within it, with specific values, and an external representation. In practice, and looking through the .scalarValue data types, I think the only one that's subject to "local interpretation" is the timestamp value, but as we advance into more "attributed strings" territory, I think the concept of a common reference document will hold us in good stead, and we can make sure that some level of "marks" or attributes/blocks are equally matched in cross-platform representations.

> **UPDATE:** opened a PR to add an interop exemplar document into Automerge core: https://github.com/automerge/automerge/pull/895

heckj:
@lightsprint09 I think your rebasing went a bit awry here, GH is complaining about it being "unmergable"

lightsprint09:
Looks good to me?

heckj:
Screenshot 2024-04-02 at 11 47 44 AM

lightsprint09:
Bildschirmfoto 2024-04-02 um 20 53 13

hmm ;)

heckj (approved):

heckj:
Odd - if you can merge it, go ahead - it's been complaining since you merge back the main branch, but if it's clear for you, awesome.

PR #153, opened 2024-04-29
Fix error when decoding int as double

bgomberg:
I found that if I create an Automerge document in Javascript with a field being defined as a `number`, assign an integer value into it, and then try to load that document in Swift, I get an error saying `Expected to decode Double but found ScalarValue<Int(5)> instead.`. I would expect the decoder to just handle casting ints to doubles, as this is what the Swift JSON decoder does. This PR updates the behavior to match that expectation.


heckj:
Thank you @bgomberg - looks great!

Issue #148, opened 2024-04-29
expose the diff set of APIs


heckj:
Wire through the the diff API sets through the UniFFI and Rust library code to expose that API into the Swift library.

Upstream docs for Rust: https://docs.rs/automerge/latest/automerge/struct.AutoCommit.html#method.diff

- `diff_cursor`
- `diff`
- `diff_incremental`

PR #155, opened 2024-04-30
0.5.10 core library update

heckj:
- update to core library version 0.5.10
- PatchAction::Insert no longer accepts marks with the 0.5.10 API
  - remove from UDL
  - remove from Rust code making library overlay
  - remove from Swift API

### DIFF
--- a/AutomergeUniffi/automerge.swift
+++ b/AutomergeUniffi/automerge.swift
@@ -642,11 +642,12 @@ open class Doc:
     }
 
     public convenience init() {
-        let pointer = try! rustCall {
-            uniffi_uniffi_automerge_fn_constructor_doc_new(
-                $0
-            )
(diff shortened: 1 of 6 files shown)


heckj:
@alexjg when you have some time, would you look over these changes - I _think_ the only notable API change was removing the marks parameter from `PatchAction::Insert`, but I may have not spotted other API updates that are important here.

heckj:
I found the relevant commits in the recent history - I think this covers the required changes for now, and opened #156 (in addition to #148 ) to expose more API through to the Swift language down the road.

alexjg:
Apologies, I missed this mention. Yes the only change is to the `Insert` patch.

heckj:
No worries, thanks for the double-check!

PR #169, opened 2024-05-07
revising core api to reference type of object in schema

### LINKED ISSUE #86: consider renaming the raw document update types to include some hints of what kind of document objects they work with
from @jessegrosjean in Discord chat:

since ObjIds are not typed, it might make API easier to read if actions are named based on type they can happen to. Right now some are, some are not. And it gets a little confusing since you have to switch on prop for some to determine what kind of object you are working with. So maybe more consistent would be action names:
- MapPut
- MapDelete
- MapConflict
- ListSplice
- ListPut
- ListConflict
- TextSplice
- TextMarks
- CounterIncrement

heckj:
# Work In Progress

Core API updates - primarily replicating and renaming original API methods to shift to a usage that has the type of internal schema object being used, to help with clarity when using the core API. 

- [ ] update documentation
- [ ] update tests to resolve all "deprecation" warnings

A classic example of what this deprecation will look like in existing code:
Screenshot 2024-05-07 at 4 13 36 PM

### DIFF
--- a/Sources/Automerge/BoundTypes/AutomergeText.swift
+++ b/Sources/Automerge/BoundTypes/AutomergeText.swift
@@ -395,7 +395,7 @@ import os
 
 extension AutomergeText: ObservableObject {
     fileprivate func sendObjectWillChange() {
-// DEBUGGING / DIAGNOSTICS CODE to show where object changes are being initiated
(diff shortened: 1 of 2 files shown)


heckj:
Further updates pending community feedback...

heckj:
After letting this sit in my head for a while, I think the better option is to close this and not make these updates.

Issue #156, opened 2024-05-01
expose Blocks API to Swift


heckj:
The core API:
- split_block
- join_block
- update_block
- spans
- updateSpans

For docs (Rust) ref: https://github.com/automerge/automerge/commit/806ef43a50dcf1038291567ddca4e4b135ecc69b

The rust commit that added most of the core: https://github.com/automerge/automerge/commit/1d987cd4c25f4d7c6de46567ace5058761532855
The commit that exposed through to WASM: https://github.com/automerge/automerge/commit/5068143802d33b84d17d335d489240d02853a74f
The commit that exposed through to JavaScript: https://github.com/automerge/automerge/commit/e7b090ae64247febd80c471cc97c9ed63d85f01d

The blocks API should allow us to provide mapping from some common JavaScript-based rich-text editors through to Swift's AttributedString type.

heckj:
Digging through existing tests, code, etc - 

Looks like the types to expose from Rust over into swift include `BlockOrText` - an enum that holds either the block structure or a string value (from automerge/tests/block_tests.rs)

The map inside block has:
- String(`type`) => String
- String(`parents`) => [String]
- String(`attrs`) => [String:?? something - ScalarValue? ints?]
  - For the attrs map, the example in the test showed a key of a string and a value of an int
  
A lot of those parts are using functions: `hydrate_map`, `hydrate_list`, `hydrate_text` that may be useful to expose and leverage, haven't read through them all yet.

The exposed [WASM tests](https://github.com/automerge/automerge/blob/main/rust/automerge-wasm/test/blocks.mts) illustrate potentially more of how this might be better exposed:

heckj:
UniFFI exposes a Hashmap as a default type (https://mozilla.github.io/uniffi-rs/udl/builtin_types.html), which it looks like we'll need to use for some of this API which accepts (and returns) maps. the `hydrate.rs` module features prominently in the api, and it's Value type (enum) has cases for `List` and `Map` in addition to `Object` and `ScalarValue`. We may want/need to expand that in the Automerge-swift Value type - or accomodate that in some form (a different type to avoid futzing with existing API?) to accomodate the more complex structures in spans and blocks

PR #171, opened 2024-05-16
updating mark API to align with core library, exposing split_block and join_block through to Swift

heckj:
Work against #156 

- [x] mark (Value -> ScalarValue)
- [x] split_block
- [x] join_block


heckj:
Constraining this PR to (primarily) the earlier marks updates to align with the updated Core library API to accomodate parallel development for some additional marks features in #186

PR #174, opened 2024-05-24
AutomergeUniffi compile target for visionos and visionos-simulator

cklokmose:
Updated to the latest rust nightly and created a compile target for visionOS.


Package.swift
```diff
@@ -1,10 +1,10 @@
-// swift-tools-version:5.6
+// swift-tools-version:5.9
```
heckj, line 1: The only possible downside here is that this constrains support to iOS 12 (per https://forums.developer.apple.com/forums/thread/734896), so earlier versions of iOS are dropped with this push forward. I don't think that'll be an issue, but we'll check with the community on Discord to be sure.

Package.swift
```diff
@@ -64,7 +64,7 @@ if ProcessInfo.processInfo.environment["LOCAL_BUILD"] != nil {
 
 let package = Package(
     name: "Automerge",
-    platforms: [.iOS(.v13), .macOS(.v10_15)],
+    platforms: [.iOS(.v13), .macOS(.v10_15), .visionOS(.v1)],
```
heckj, line 67: I don't think this is strictly needed, as this is the _base_ version of the OS that the library will work with, and .v1 was the first version ever, but it also doesn't hurt anything.

scripts/build-xcframework.sh
```diff
@@ -32,7 +32,7 @@ XCFRAMEWORK_FOLDER="$THIS_SCRIPT_DIR/../${FRAMEWORK_NAME}.xcframework"
 # (as of 10/10/23), but leaving it open to float seems less useful than
 # moving the pinning forward, since Catalyst support (target macabi) still
 # requires an active, nightly toolchain.
-RUST_NIGHTLY="nightly-2023-10-09"
+RUST_NIGHTLY="nightly-2024-05-23"
```
heckj, line 35: There's a cooresponding version of Rust that's matched with this nightly toolchain that we want to make sure is aligned (https://github.com/automerge/automerge-swift/blob/main/.github/workflows/ci.yaml#L18, and a few other places). In CI, we're running Rust checks with 1.75.0 - does that need to be pushed forward to align with this?

heckj (commented):
I'd like to check with the community about the iOS support potential, and would like to know about Rust versions to align CI correctly with the nightly you choose.

scripts/build-xcframework.sh
cklokmose: For visionos the toolchain is compiled. We could also consider using the newest nightly only for the visionos target for now?

Package.swift
cklokmose: I believe 5.9 is needed to include visionos as a target.

heckj: The WASM build also crapped out with this, and dropping WASM support in favor of VisionOS isn't a good move, so I think we'll need to sort how how to enable that as well.. We'll need to dig into https://github.com/automerge/automerge-swift/blob/main/.github/workflows/ci.yaml#L43 and see if we can nail down a WASM toolchain for Swift 5.9 to continue that.

cklokmose: Oh! That's a good point.

scripts/build-xcframework.sh
heckj: Looks like the nightly you picked is after 1.78, but the core bits not yet nailed down into 1.79. I think pushing CI to use 1.78 would be the first steps, and we can keep using the nightly where it stands.

Package.swift
heckj: I opened #175 to update the CI toolchains indepdent of this PR, so that if that goes smoothly, we can merge it and have this try again and see how CI fares. I didn't investigate any of the [swift build failures](https://github.com/automerge/automerge-swift/actions/runs/9228143274/job/25392987557?pr=174) though

PR #183, opened 2024-06-06
exposing Automerge::Diff through to Swift

miguelangel-dev:
Bindings for being able to compare two points of the document history.

- #184

Related documentation: https://docs.rs/automerge/latest/automerge/struct.Automerge.html#method.diff

### DIFF
--- a/Sources/Automerge/Automerge.docc/Curation/Document.md
+++ b/Sources/Automerge/Automerge.docc/Curation/Document.md
@@ -77,6 +77,7 @@
 - ``change(hash:)``
 - ``difference(from:to:)``
 - ``difference(from:)``
+- ``difference(to:)``
 
 ### Reading historical map values
 
--- a/Sources/Automerge/Automerge.docc/Curation/Document.md
+++ b/Sources/Automerge/Automerge.docc/Curation/Document.md
@@ -75,6 +75,8 @@
 - ``heads()``
 - ``getHistory()``
 - ``change(hash:)``
+- ``difference(from:to:)``
+- ``difference(from:)``
 
 ### Reading historical map values
 
--- a/Sources/Automerge/Automerge.docc/Curation/Document.md
+++ b/Sources/Automerge/Automerge.docc/Curation/Document.md
@@ -76,7 +76,7 @@
 - ``getHistory()``
 - ``change(hash:)``
 - ``difference(from:to:)``
-- ``difference(from:)``
+- ``difference(since:)``
 - ``difference(to:)``
 
 ### Reading historical map values
--- a/Sources/Automerge/Document.swift
+++ b/Sources/Automerge/Document.swift
@@ -954,7 +954,7 @@ public final class Document: @unchecked Sendable {
     /// doc.difference(to: doc.heads())
     /// ```
     /// - Parameters:
-    ///   - from: heads at beginning point in the documents history.
+    ///   - from: The set of heads at beginning point in the documents history.
     ///   - to: heads at ending point in the documents history.
     /// - Note: `from` and `to` do not have to be chronological. Document state can move backward.
     /// - Returns: The difference needed to produce a document at `to` when it is set at `from` in history.
--- a/Sources/Automerge/Document.swift
+++ b/Sources/Automerge/Document.swift
@@ -946,6 +946,27 @@ public final class Document: @unchecked Sendable {
         }
     }
 
+    /// Generates patches between two points in the document history.
+    ///
+    /// Use:
+    /// ```
+    /// let doc = Document()
+    /// doc.difference(to: doc.heads())
+    /// ```
+    /// - Parameters:
+    ///   - from: heads at beginning point in the documents history.
+    ///   - to: heads at ending point in the documents history.
+    /// - Note: `from` and `to` do not have to be chronological. Document state can move backward.
+    /// - Returns: The difference needed to produce a document at `to` when it is set at `from` in history.
+    public func difference(from before: Set<ChangeHash>, to after: Set<ChangeHash>) -> [Patch] {
+        sync {
+            let patches = self.doc.wrapErrors { doc in
+                doc.difference(before: before.map(\.bytes), after: after.map(\.bytes))
+            }
+            return patches.map { Patch($0) }
+        }
+    }
+
     /// Get the path to an object within the document.
     ///
     /// - Parameter obj: The identifier of an array, dictionary or text object.
--- a/Sources/Automerge/Document.swift
+++ b/Sources/Automerge/Document.swift
@@ -967,6 +967,37 @@ public final class Document: @unchecked Sendable {
         }
     }
 
+    /// Generates patches **from** a given point in the document history.
+    ///
+    /// Use:
+    /// ```
+    /// let doc = Document()
+    /// doc.difference(from: doc.heads())
+    /// ```
+    /// - Parameters:
+    ///     - from: heads at beginning point in the documents history.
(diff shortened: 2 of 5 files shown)


Sources/Automerge/Document.swift
```diff
@@ -946,6 +946,58 @@ public final class Document: @unchecked Sendable {
...
+    /// Use:
+    /// ```
+    /// let doc = Document()
+    /// doc.difference(to: doc.heads())
+    /// ```
+    /// - Parameters:
+    ///     - to: heads at ending point in the documents history.
+    /// - Note: `from` do not have to be chronological. Document state can move backward.
+    /// - Returns: The difference needed to produce a document at `to` in the history.
+    public func difference(to rhs: Set<ChangeHash>) -> [Patch] {
```
heckj, line 1005: If the order of the diff doesn't make any difference, then only one of the single parameter versions is really needed. By having two, the API is re-inforcing that there's a strict ordering needed. I'd recommend removing one of these.

Sources/Automerge/Document.swift
```diff
@@ -946,6 +946,58 @@ public final class Document: @unchecked Sendable {
...
+
+    /// Generates patches **from** a given point in the document history.
+    ///
+    /// Use:
+    /// ```
+    /// let doc = Document()
+    /// doc.difference(from: doc.heads())
+    /// ```
+    /// - Parameters:
+    ///     - from: heads at beginning point in the documents history.
```
```suggestion
    ///     - since: The set of heads at the point in the documents history to compare to.
```

Sources/Automerge/Document.swift
```diff
@@ -946,6 +946,58 @@ public final class Document: @unchecked Sendable {
...
 
+    /// Generates patches between two points in the document history.
+    ///
+    /// Use:
+    /// ```
+    /// let doc = Document()
+    /// doc.difference(to: doc.heads())
+    /// ```
+    /// - Parameters:
+    ///   - from: heads at beginning point in the documents history.
```
```suggestion
    ///   - from: The set of heads at beginning point in the documents history.
```

Sources/Automerge/Document.swift
```diff
@@ -946,6 +946,58 @@ public final class Document: @unchecked Sendable {
...
+    ///   - to: heads at ending point in the documents history.
```
```suggestion
    ///   - to: The set of heads at ending point in the documents history.
```

Sources/Automerge/Document.swift
```diff
@@ -946,6 +946,58 @@ public final class Document: @unchecked Sendable {
...
+    /// Generates patches **from** a given point in the document history.
+    ///
+    /// Use:
+    /// ```
+    /// let doc = Document()
+    /// doc.difference(from: doc.heads())
+    /// ```
+    /// - Parameters:
+    ///     - from: heads at beginning point in the documents history.
+    /// - Note: `from` do not have to be chronological. Document state can move backward.
```
```suggestion
```

Sources/Automerge/Document.swift
```diff
@@ -946,6 +946,58 @@ public final class Document: @unchecked Sendable {
...
+    /// - Returns: The difference needed to produce current document given an arbitrary
+    /// point in the history.
+    public func difference(from lhs: Set<ChangeHash>) -> [Patch] {
```
```suggestion
    public func difference(since lhs: Set<ChangeHash>) -> [Patch] {
```

Using the parameter name `from` or `to` implies directional ordering, and since that is irrelevant (I'm basing that on your comment here) and one side of this function is always the current heads(), that implies this is a convenience function to look back to a prior state. Along those lines `since` distinguishes it as such, implying more of what its doing. Along those lines, I'd drop the replica of this function that reverses the call order if that really is irrelevant.

Sources/Automerge/Automerge.docc/Curation/Document.md
```diff
@@ -75,6 +75,9 @@
 - ``heads()``
 - ``getHistory()``
 - ``change(hash:)``
+- ``difference(from:to:)``
+- ``difference(from:)``
```
heckj:
```suggestion
- ``difference(since:)``
```

Sources/Automerge/Automerge.docc/Curation/Document.md
```diff
@@ -75,6 +75,9 @@
...
+- ``difference(to:)``
```
heckj, line 80:
```suggestion
```

heckj (changes requested):
Generally this looks great, thank you!

We don't need both difference(from:) and difference(to:) if they're effectively identical APIs with just slightly different parameters. Drop one, and I've made suggestions to change the parameter since to `since` since the other side that's passed to the core library's `difference` is the current set of heads. 

I'm a unclear on why the WASM build didn't appear to pick up the change. Visually it looks correct to me, but we might need to dig a bit further. We did just update the CI runners to an updated version of Rust, which may have had a follow on effect that's hitting us here.

miguelangel-dev:
> ... if they're effectively identical APIs with just slightly different parameters. Drop one, and I've made suggestions to change the parameter since to `since` since the other side that's passed to the core library's `difference` is the current set of heads.

Probably the API doc is not the best one and this note has confused you:
```
`from` and `to` do not have to be chronological. Document state can move backward.
```

Saying both do not need to be chronological (having order) does not mean the result won't change. Here is an example: let's say I append "World" to the existing "Hello"
- if `difference` is applied in chronological order, then operations will be "inserts"
- If `difference` is applied in reverse order, operations will be "removals"

Any suggestion for the DocC, to make it more clear is more than welcome 😅

> I'm a unclear on why the WASM build didn't appear to pick up the change...

Not sure, but it looks like CI is always compiling against the latest XCFramework, instead of building a local one and using it.

Screenshot 2024-06-06 at 19 26 48

heckj:
I saw the same thing in the logs re: downloading the XCFramework. It shouldn't be a requirement for builds on Linux or WASI, but I'd agree that this might be the bit that's hurting us. Unfortunately, XCFramework - while downloaded - can't be rebuilt on just Linux, so we don't have the same path to using a "local version".

If you've tried this on your own build servers and it works - flows through WASM builds correctly, etc - then I'm happy to merge this with that check breaking, on the idea that we can pop a release with the Rust updates which theoretically _should_ resolve the issue for future builds after there's an updated point release.

miguelangel-dev:
> If you've tried this on your own build servers and it works - flows through WASM builds correctly, etc - then I'm happy to merge this with that check breaking, on the idea that we can pop a release with the Rust updates which theoretically _should_ resolve the issue for future builds after there's an updated point release.

Adding manually missing autogenerate code (headers + automerge.swift), WASM is working 🎉. We should update the Linux bot to generate it automatically without committing it - please, double-check [868e8cd](https://github.com/automerge/automerge-swift/pull/183/commits/868e8cd146a538a09701b7486dec4389e8fe1c27) because it contains more changes than I expected. Probably, because of my rust version (nightly-2024-05-23-aarch64-apple-darwin)

**side-note**: I can revert this latest commit with the autogenerated code.

heckj (approved):
I got the CI stuff updated - good call - with https://github.com/automerge/automerge-swift/pull/185, and merged that in. You can rebase if you want to double check that this PR works without the checked in updates, but don't feel obliged. 

Thanks for clarifying that directionality IS reflected based on the two sets of heads provided to difference, makes a lot more sense to have the various API

miguelangel-dev:
> I got the CI stuff updated - good call - with #185, and merged that in. You can rebase if you want to double check that this PR works without the checked in updates, but don't feel obliged.

Awesome! I have reverted the commits with the auto-generated code, to keep the PR clean and have rebased changes into this branch. Result: it works like a charm! 

Thanks for the parallel work done into #185 - it is now ready to be merged

PR #186, opened 2024-06-11
exposing Automerge::get_marks through to Swift 

miguelangel-dev:
Bindings for being able to retrieve the marks of a given position without traversing the ops of the full document.

- #187

https://automerge.org/automerge/automerge/struct.Automerge.html#method.get_marks


heckj:
Thanks @miguelangel-dev, I'll take a look in depth a little later today or tomorrow, WWDC is dominating my attention right now. In the meantime, Marks has some underlying changes that we should get in before dropping this in place - some API updates that align to the latest core library that I have in some commits in the WIP Blocks PR (#171). I'll pull those out of the existing PR and get those set up so we can merge them without waiting for the full detail work on the Blocks API

heckj:
@miguelangel-dev I don't think there'll be a serious conflict, but go ahead and rebase over the #171 merge, and things should be good. Biggest detail change here was the internal API switching from Value to ScalarValue inside marks, which I wanted to get in if we were expanding the marks API with these bits.

miguelangel-dev:
> @miguelangel-dev I don't think there'll be a serious conflict, but go ahead and rebase over the #171 merge, and things should be good. Biggest detail change here was the internal API switching from Value to ScalarValue inside marks, which I wanted to get in if we were expanding the marks API with these bits.

Thanks for moving and merging. I have rebased it against #171 - it should be ready to review 🙏🏻

rust/src/doc.rs
```diff
@@ -481,6 +476,43 @@ impl Doc {
             .collect())
     }
 
+    pub fn marks_cursor(
```
alexjg: Given that this and `marks_position` are so similar I would suggest unifying the definitions with an enum over the position. Something like this:

```rust
enum Position {
    Cursor(am::Cursor),
    Index(usize),
}


pub fn marks_at_position(&self, obj: ObjId, position: Position, heads: Vec<ChangeHash>) { .. }
```

rust/src/mark.rs
```diff
@@ -37,3 +37,24 @@ impl<'a> From<&'a am::marks::Mark<'a>> for Mark {
         }
     }
 }
+
+pub struct KeyValue {
+    pub key: String,
+    pub value: Value,
```
heckj, line 43:
```suggestion
    pub value: ScalarValue,
```

I'm reasonable certain that this should be `ScalarValue`, not the broader `Value` type, to align with the marks API updates. It used to return `Value`, but that was part of the refactoring bit - it's limited to `ScalarValue`.

rust/src/mark.rs
```diff
@@ -1,6 +1,6 @@
 use automerge as am;
 
-use crate::ScalarValue;
+use crate::{ScalarValue, Value};
```
heckj, line 3:
```suggestion
use crate::ScalarValue;
```

With `KeyValue` being a struct over `ScalarValue` instead of `Value`, I don't think you'll need this inclusion.

Sources/Automerge/Document.swift
```diff
@@ -768,6 +768,61 @@ public final class Document: @unchecked Sendable {
...
+    /// let textId = try doc.putObject(obj: ObjId.ROOT, key: "text", ty: .Text)
+    ///
+    /// let cursor = try doc.cursor(obj: textId, position: 0)
+    /// let marks = try doc.marksAt(obj: textId, cursor: cursor)
+    /// ```
+    ///
+    /// - Parameters:
+    ///   - obj: The identifier of the text object.
+    ///   - cursor: The cursor that indicates a position within the text.
+    ///   - heads: The set of ``ChangeHash`` that represents a point of time in the history the document.
```
heckj: Since this is set up as optional, include some detail about what the default operation does when the value is not provided by the developer (e.g. "uses the latest version of the document")

Sources/Automerge/Document.swift
```diff
@@ -768,6 +768,61 @@ public final class Document: @unchecked Sendable {
...
+    /// let doc = Document()
+    /// let textId = try doc.putObject(obj: ObjId.ROOT, key: "text", ty: .Text)
+    ///
+    /// let marks = try doc.marksAt(obj: textId, position: 0)
+    /// ```
+    ///
+    /// - Parameters:
+    ///   - obj: The identifier of the text object.
+    ///   - position: The index that indicates the position of the text offset.
+    ///   - heads: The set of ``ChangeHash`` that represents a point of time in the history the document.
```
heckj: Since this is set up as optional, include some detail about what the default operation does when the value is not provided by the developer (e.g. "uses the latest version of the document")

Sources/Automerge/Document.swift
```diff
@@ -768,6 +768,61 @@ public final class Document: @unchecked Sendable {
         }
     }
 
+    /// Get the list of marks for a text object at ``Cursor``.
```
heckj: Using ``Cursor`` symbol as the first line (the abstract) links in an awkward way in the DocC output. I'd recommend something more like:

```suggestion
    /// Get the list of marks inside a text object at the position you indicate with a cursor.
```

And link to cursor in the content of the reference, if you want to do that.

Sources/Automerge/Document.swift
```diff
@@ -768,6 +768,61 @@ public final class Document: @unchecked Sendable {
...
+    ///
+    /// Use:
```
heckj: Since this is so nearly identical to `MarksAt(obj:position:heads)`, it would be good to include some detail on why a developer should use this method compared to a straight index position. The example above shows the how, but is identifical to the position example in that nothing is changing in the Text object that the cursor would properly track in terms of location over time.

rust/src/mark.rs
miguelangel-dev: It would be nice, but this existing structure has a more general purpose. It is also used for `map_entries` where `Value` is mandatory.

rust/src/mark.rs
miguelangel-dev: Please, double check this comment: https://github.com/automerge/automerge-swift/pull/186#discussion_r1636963716 - it seems `map_entries` need to work over `Value`

Sources/Automerge/Document.swift
miguelangel-dev: It was a bit lazy from my side, better to split the method into two variants.

Sources/Automerge/Document.swift
miguelangel-dev: Oh good call - I will do

rust/src/doc.rs
miguelangel-dev: Suggestion addressed in 55763e3897fd1273a50e5eb1f1093aaab284b4de

Sources/Automerge/Document.swift
miguelangel-dev: Improved the documentation in 63070af2a54f944c6b831143d0d11c481ba3d27f - introducing extra information of differences between using `Cursor` and `Index.

Sources/Automerge/Document.swift
miguelangel-dev: https://github.com/automerge/automerge-swift/pull/186#discussion_r1637157757

Sources/Automerge/Document.swift
miguelangel-dev: https://github.com/automerge/automerge-swift/pull/186#discussion_r1637157757

miguelangel-dev:
Thanks both for the code review, I have addressed all your suggestions. It is ready for a second round 🙏🏻

heckj (approved):

miguelangel-dev:
Sorry, added a latest minor change to revert an unnecessary public accessibility into the cursor c6798ffe4a6dacb43abe8ee6576c60345138c7eb

alexjg (approved):

heckj (approved):

PR #190, opened 2024-07-10
Encode `nil` values into the document

marionauta:
The encoder was omitting some `nil` values, making them `undefined` in the JavaScript library, instead of `null`.

Given the following code:

```swift
struct Item: Codable {
    var completedAt: Date?

    func encode(to encoder: any Encoder) throws {
        var container = encoder.container(keyedBy: CodingKeys.self)
        try container.encode(completedAt, forKey: .completedAt)
    }
}

let item = Item(completedAt: nil)
let doc = Document()
let encoder = AutomergeEncoder(doc: doc)
try encoder.encode(item)
let completedAt = try doc.get(obj: .ROOT, key: "completedAt")
print(completedAt)
```

Results in `completedAt` being `nil`, instead of `ScalarValue.Null`. This means the encoder wasn't writing the null value into the `Document`.

---

This PR changes the encoder so it writes `nil`s into the `Document`, with tests to check it.

As an implementation detail, now `ScalarValue` conforms to `ScalarValueRepresentable`.

### DIFF
--- a/Sources/Automerge/Codable/Encoding/AutomergeSingleValueEncodingContainer.swift
+++ b/Sources/Automerge/Codable/Encoding/AutomergeSingleValueEncodingContainer.swift
@@ -55,7 +55,10 @@ struct AutomergeSingleValueEncodingContainer: SingleValueEncodingContainer {
         #endif
     }
 
-    mutating func encodeNil() throws {}
+    mutating func encodeNil() throws {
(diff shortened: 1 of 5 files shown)


heckj (approved):
Looks great, Mario! Thank you for the detailed tests hitting not just a nil parameter in a map, but an array of potentially nil items as well.

PR #191, opened 2024-07-11
Expose heads raw state representation

miguelangel-dev:
### Motivation
Sometimes we want to persist in our domain a state representation based on `Set<ChangeHash>` (aka heads) without coupling to the Automerge library. In this way, we can restore this state by moving back to the Automerge domain. It is usually quite interesting to implement an undo manager based on "commits".

Introducing a `raw()` method to retrieve the raw representation of `heads` without any Automerge dependency; and also the go-back action into `Set<ChangeHash>`.

side-note: as we use an opaque type to encode the raw version of a group of change hashes, I am just checking it contains packages of 32 bytes to retrieve the associated change hash - after retrieving it, in case of not valid ChangeHash document will throw exceptions.

### DIFF
--- a/Sources/Automerge/ChangeHash.swift
+++ b/Sources/Automerge/ChangeHash.swift
@@ -18,7 +18,7 @@ public extension Set<ChangeHash> {
     /// and retrieval of document states.
     func raw() -> Data {
         let rawBytes = map(\.bytes).sorted { lhs, rhs in
-            lhs.hashValue > rhs.hashValue
+            lhs.debugDescription > rhs.debugDescription
         }
         return Data(rawBytes.joined())
     }
@@ -28,7 +28,7 @@ public extension Data {
 
     /// Returns the related set of changes of a state representation within an Automerge document.
     func heads() -> Set<ChangeHash>? {
-        let rawBytes = Array(self)
+        let rawBytes: [UInt8] = Array(self)
         guard rawBytes.count % 32 == 0 else { return nil }
         let totalHashes = rawBytes.count / 32
         let heads = (0..<totalHashes).map { index in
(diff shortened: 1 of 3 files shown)


Sources/Automerge/ChangeHash.swift
```diff
@@ -9,3 +10,33 @@ public struct ChangeHash: Equatable, Hashable, CustomDebugStringConvertible, Sen
...
+        let rawBytes = map(\.bytes).sorted { lhs, rhs in
+            lhs.debugDescription > rhs.debugDescription
+        }
+        return Data(rawBytes.joined())
+    }
+}
+
+public extension Data {
+
+    /// Returns the related set of changes of a state representation within an Automerge document.
```
heckj:
```suggestion
    /// Interprets the data to return the data as a set of change hashes that represent a state within an Automerge document. If the data is not a multiple of 32 bytes, returns nil.
```

heckj (approved):
Since the relevant methods are kind of hidden on `Set<ChangeHash>` and `Data`, I think it's worth adding something to the overall documentation for ChangeHash, or potentially `Document.heads()` to reference them  - otherwise they'll be really hard to find for anyone searching for this functionality. Not critical for this PR though.

alexjg:
One thing that might be worth thinking about. In the `automerge` crate we have a `TryFrom<&[u8]>` implementation for `ChangeHash`. This is the supported way of converting bytes into a `ChangeHash` and is where we would handle future inconsistencies in hash format (not that I expect any at this point). If possible I think it would be best to expose this method up through the Uniffi UDL rather than assuming that any 32 byte array is a valid change hash. (This also applies to the serialization step, except we're already using `ChangeHash::as_bytes` when we generate the bytes which we pass to the Swift side).

heckj:
Created #193 to extend this to expose, and use, the tryFrom logic in the Rust core library within the internal initializer for ChangeHash

PR #192, opened 2024-07-13
Protect document state with recursive lock and add objectDidChange publisher

### LINKED ISSUE #178: Add doc.objectDidChange?
In addition to `objectWillChange` I often feel myself wanting an `objectDidChange` event. I know SwiftUI doesn't need it, but generally didChange events seems like the useful default. Does it make sense to add this event to doc. And also fire it outside the sync block so that document state can be read immediately when receiving the event?

jessegrosjean:
This gives Document an .objectDidChange publisher, that is fired in sync with and immediately after changes are made to a document. This required two changes:

1. Protect document state with a NSRecursiveLock instead of a dispatch queue. This is needed because I want to be able to call back to the document from my .objectDidChange listener without creating deadlock in the queue.
2. Add an optional (when combine is available) .objectDidChange publisher that always balances the current .objectWillChange publisher.

resolves #178

### DIFF
--- a/Sources/Automerge/Document.swift
+++ b/Sources/Automerge/Document.swift
@@ -16,12 +16,14 @@ public final class Document: @unchecked Sendable {
     private var doc: WrappedDoc
 
     #if !os(WASI)
-    fileprivate let queue = DispatchQueue(label: "automerge-sync-queue", qos: .userInteractive)
-    fileprivate func sync<T>(execute work: () throws -> T) rethrows -> T {
-        try queue.sync(execute: work)
+    let lock = NSRecursiveLock()
+    fileprivate func lock<T>(execute work: () throws -> T) rethrows -> T {
+        lock.lock()
+        defer { lock.unlock() }
+        return try work()
     }
     #else
-    fileprivate func sync<T>(execute work: () throws -> T) rethrows -> T {
+    fileprivate func lock<T>(execute work: () throws -> T) rethrows -> T {
         try work()
     }
     #endif
@@ -31,12 +33,12 @@ public final class Document: @unchecked Sendable {
     /// The actor ID of this document
     public var actor: ActorId {
         get {
-            sync {
+            lock {
                 ActorId(ffi: self.doc.wrapErrors { $0.actorId() })
             }
         }
         set {
-            sync {
+            lock {
                 self.doc.wrapErrors { $0.setActor(actor: [UInt8](newValue.data)) }
             }
         }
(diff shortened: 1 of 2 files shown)


Sources/Automerge/Document.swift
```diff
@@ -16,27 +16,33 @@ public final class Document: @unchecked Sendable {
...
     }
     #else
-    fileprivate func sync<T>(execute work: () throws -> T) rethrows -> T {
+    fileprivate func lock<T>(execute work: () throws -> T) rethrows -> T {
         try work()
     }
     #endif
+    
+    #if canImport(Combine)
+    public let objectDidChange: PassthroughSubject<(), Never> = .init()
```
heckj, line 32: Can you add a bit of doc comment here highlighting that this is triggered explicitly _after_ the document has been updated, and use your use-case as a "for example"? i.e. "Sends a signal that the document has been updated. For example, you can use this signal to retrieve `heads` for the document to capture the state after you applied changes." - or something like that?

heckj (changes requested):

Sources/Automerge/Document.swift
jessegrosjean: Yes, but I'm just in the middle of a mad summer dash, so won't get to it today, but will do.

alexjg:
I _think_ this is all good. I have a slightly uncomfortable feeling about the re-entrancy because we use an `RwLock` on the rust side to ensure exclusive access to the document and we panic if we can't obtain the lock because we know that on the Swift side we're ensuring exclusive access. However, I think the re-entrancy here would never happen when we're holding the lock on the Rust side so we're probably okay.

To make that concrete, here's `put_in_map`


```rust
    pub fn put_in_map(&self, obj: ObjId, key: String, value: ScalarValue) -> Result<(), DocError> {
        let obj = am::ObjId::from(obj);
        let mut doc = self.0.write().unwrap();
        assert_map(&*doc, &obj)?;
        doc.put(obj, key, value).map_err(|e| e.into())
    }
```

(`self.0` is an `RwLock<automerge::AutoCommit>`)

You can see here that we panic if we can't obtain the lock (the `self.0.write().unwrap()` call). Now, none of the methods on `Document` return anything which refers to the document - and I don't think there is any way we could return such a reference as Uniffi won't let us - so we're okay. So, despite my uneasiness, I think this works.

alexjg (approved):

Sources/Automerge/Document.swift
heckj: Jesse - I'll go ahead and merge this and add the docs after.

heckj:
On the Apple platforms side, NSRecursiveLock (the lock in question here) **only** allows the same thread to pass through the lock, so I think it prevents the bits where we'd get re-entrancy across multiple threads, and with a single thread access we'd be fine - it would come through to the Rust side as serialized calls to the llibrary since they're all coming from a single thread. Other lock implementations here could be very poor, so I'd want to not replace this with a "cheaper lock" down the road, as the different semantics could easily bite us, but this I think is going to be fine - a good choice for the pain point/use-case that Jesse's hitting.

PR #195, opened 2024-07-21
Document object did change

jessegrosjean:
Sorry this is late, just getting back to computer work now. I see you've already added some docs to objectDidChange, but this adds a little more detail and an example. It also makes the publishing end of objectDidChange private.


heckj:
Nothing is late when you're volunteering work. Thank you for the tweak to make it AnyPublisher and the doc updates!

PR #198, opened 2024-10-16
Prevents coupling UTF-16 text value encodings to WASM targets

miguelangel-dev:
### Related PRs
- https://github.com/automerge/automerge/pull/977

### Description
Because of misalignments in encoding width between UTF-16 and Unicode Scalars; when we use `emoji`, or any other Unicode character, splice-text, markers and cursors won't work.

#### Using Automerge 0.5.11
Previous to apply [automerge#977](https://github.com/automerge/automerge/pull/977)
Screenshot 2024-10-21 at 16 32 00

#### Using Automerge 0.5.12
Screenshot 2024-10-21 at 16 53 47


.github/workflows/ci.yaml
```diff
@@ -29,26 +29,43 @@ jobs:
       - name: Swift tests
         run: swift test
 
-  build-wasm32-unknown-wasi:
+  wasm:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - toolchain: swift-DEVELOPMENT-SNAPSHOT-2024-10-08-a
```
heckj, line 78: Is there a release toolchain that we can use here, and not rely on a development snapshot chain?

.github/workflows/ci.yaml
```diff
@@ -29,26 +29,43 @@ jobs:
...
           profile: minimal
-          toolchain: 1.78.0
+          toolchain: 1.80.0
           default: true
-      - run: rustup target add wasm32-wasi
-      - uses: swiftwasm/setup-swiftwasm@v1
-        with:
-          swift-version: "wasm-5.9.1-RELEASE"
-      #- run: cargo build --manifest-path rust/Cargo.toml --target wasm32-wasi --release
+      - run: rustup target add wasm32-wasip1-threads
```
heckj: I don't know a huge amount about the specifics, but recognize we're changing the target platform here from `wasm32-wasi` to `wasm32-wasip1-threads`. I'm guessing this will need to evolve down the road as WASM evolves as well - most of the (current) swift WASM generation (Swift 5.8 -> Swift 6) works with wasi preview 1 specifically, but I don't know the differences involved otherwise. I think GoodNotes are the only folks using this, but are there notes that should be made for anyone else wanting to leverage this result that would shed specifics of the WASI platform interactions here?

heckj (commented):
I'd very much prefer to NOT rely on a nightly build of anything if we can get away with it, especially for ongoing CI, but if that's the only version with a release version not yet available that does what's needed here, I can live it it for now.

.github/workflows/ci.yaml
miguelangel-dev: Providing context: `wasm32-wasi` has been renamed to `wasm32-wasip1` [May'24], and it will be deprecated next January [Jan'25] [[ref](https://blog.rust-lang.org/2024/04/09/updates-to-rusts-wasi-targets.html#renaming-wasm32-wasi-to-wasm32-wasip1)]. `wasm32-wasip1-threads` is an extension of `wasm32-wasip1` adding native multi-threading capabilities. Mainly it supports std fully, [bulk memory](https://github.com/WebAssembly/spec/blob/main/proposals/bulk-memory-operations/Overview.md), atomics, expose pthread APIs. Both are Tiers2 in Rust, and for `wasm32-wasip1-threads` it will use atomic operations for sharable resources.

Said that we can not mix objects compiled between `wasm32-wasip1-threads` and `wasm32-wasip1`; so, I think it makes sense to generate a `libuniffi_automerge` with both, and let the users pick the correct one.

.github/workflows/ci.yaml
miguelangel-dev: The current swift-wasm project, even in the latest release ([Swift 5.10](https://blog.swiftwasm.org/posts/5-10-released/)), was still tied to a development snapshot. I’ve reviewed the latest wasm updates, which include significant changes since June, such as support for wasip1-thread. Given that we don’t yet have a new official release, I believe it’s beneficial for the project to ensure compatibility with the latest development version of the wasm SDK.

Once a new official SDK is released, I will open a PR to update accordingly.

heckj: works for me

heckj (approved):
Looks completely reasonable to me. Thanks for the notes on the different platform options - I expect this to continue to be in flux over the coming few years, and hopefully there'll be options to make installation of SDKs and toolchains for CI easier (and more consistent) down the road.

PR #199, opened 2024-10-29
Swift6 mode

heckj:
# HOLD MERGE

Hold the merge until a new version of UniFFI is released to resolve the code generation issue. The PR I submitted to fix the issue we're seeing has been accepted, but further PRs are underway for a more complete UniFFI solution to generating safe Swift6 code.

(We could merge this as is, but then anyone regenerating the code with UniFFI would need to manually update that same `var` to a `let`, which doesn't make sense)

Work in progress to enable full compilation for Automerge-swift under Swift 6 language mode.

Most of this is good and ready to go, but there's a manual change I needed to make in `AutomergeUniffi/automerge.swift`, which is generated by UniFFI, that needs to be resolved before we can easily make this claim.

### DIFF
--- a/AutomergeUniffi/automerge.swift
+++ b/AutomergeUniffi/automerge.swift
@@ -11,7 +11,7 @@ import Foundation
 import automergeFFI
 #endif
 
-fileprivate extension RustBuffer {
+private extension RustBuffer {
     // Allocate a new buffer, copying the contents of a `UInt8` array.
     init(bytes: [UInt8]) {
         let rbuf = bytes.withUnsafeBufferPointer { ptr in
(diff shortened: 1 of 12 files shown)


heckj:
Full resolution will require that a new version of UniFFI is available, with the patch submitted that renders global variables as `let` instead of `var`: https://github.com/mozilla/uniffi-rs/pull/2290

heckj:
Following up on the [uniffi-rs](https://github.com/mozilla/uniffi-rs) repo and release flow, it looks like this will need to wait for the next large-ish release (`0.29.x`) as the required upstream changes didn't flow into the `0.28.x` release branch - instead they're on the `main`

PR #205, opened 2024-12-23
Mutation abstraction

lightsprint09:
This introduces some internal abstraction for executing reads & write into a document. This helps to have a singe point where to implement event handling & error conversation. At the same time concrete implementations get more clear.

### DIFF
--- a/Sources/Automerge/Document.swift
+++ b/Sources/Automerge/Document.swift
@@ -109,7 +109,7 @@ public final class Document: @unchecked Sendable {
     /// amount,
     /// use the method ``increment(obj:key:by:)`` instead.
     public func put(obj: ObjId, key: String, value: ScalarValue) throws {
-        try mutate {
+        try write {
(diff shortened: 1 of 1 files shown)


Sources/Automerge/Document.swift
```diff
@@ -58,13 +58,13 @@ public final class Document: @unchecked Sendable {
...
-            lock {
-                ActorId(ffi: self.doc.wrapErrors { $0.actorId() })
+            read {
+                ActorId(ffi: doc.actorId() )
             }
         }
         set {
-            lock {
-                self.doc.wrapErrors { $0.setActor(actor: [UInt8](newValue.data)) }
+            write {
```
lightsprint09, line 66: Should setting an Acor trigger change events?

heckj: If it's changing the underlying document structure and content in any way, then yes.

lightsprint09: Ok, so then this a missing change event when setting the actor

PR #206, opened 2025-01-20
swift: allow specifying text encoding at document creation

miguelangel-dev:
### Related pull-requests:
- Introduces text encoding: https://github.com/automerge/automerge/pull/997

### What is the goal?
Exposes new Automerge TextEncoding flag for creation of documents. It enables us to define the text representation to use in each of the operations without being forced to use `Unicode Scalars` in Swift.

**blockers:** pending to merge related PRs, and bump automerge core.


Sources/Automerge/Document.swift
```diff
@@ -658,23 +667,21 @@ public final class Document: @unchecked Sendable {
...
-    ///   - delete: The number of unicode scalars to delete from the `start` index.
+    ///   - start: The distance from the start of the string where the function begins inserting or deleting.
+    ///   - delete: Text length to delete from the `start` index. It depends on the picked ``TextEncoding`` in Document creation.
     ///   If negative, the function deletes characters preceding `start` index, rather than following it.
     ///   - value: The characters to insert after the `start` index.
     ///
-    /// With `spliceText`, the `start` and `delete` parameters represent integer distances of unicode scalars of the
-    /// Swift strings, not the counts of Characters (or grapheme clusters).
+    /// With `spliceText`, the `start` and `delete` parameters represent integer distances of the Swift strings. This distance will change
+    /// depends on the picked text representation in the Document creation.
```
```suggestion
    /// based on the text encoding chosen during at Document creation.
```

Sources/Automerge/Document.swift
```diff
@@ -658,23 +667,21 @@ public final class Document: @unchecked Sendable {
...
     ///
-    /// If you use or receive a Swift `String.Index` convert it to an index position usable by Automerge through
-    /// `UnicodeScalarView`, accessible through the `unicodeScalars` property on the string.
-    /// To determine Automerge index position from a `String.Index`, convert the index position into a
-    /// `String.UnicodeScalarView.Index` and calculate the distance from the `startIndex` value.
+    /// If you use or receive a Swift `String.Index` convert it to an index position usable by Automerge through `Foundation.String.View`
+    /// APIs. Indices depends on picked ``TextEncoding`` during Automerge.Document creation.
```
```suggestion
    /// APIs. Indices in Automerge are based on the ``TextEncoding`` chosen at document creation.
```

rust/src/text_encoding.rs
```diff
@@ -0,0 +1,42 @@
...
+            am::TextEncoding::GraphemeCluster => TextEncoding::GraphemeCluster,
+        }
+    }
+}
+
+impl std::fmt::Display for TextEncoding {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::UnicodeCodePoint => write!(f, "UnicodeCodePoint"),
+            Self::Utf8CodeUnit => write!(f, "Utf8CodeUnit"),
```
heckj:
```suggestion
            Self::Utf8CodeUnit => write!(f, "UTF-8CodeUnit"),
```

rust/src/text_encoding.rs
```diff
@@ -0,0 +1,42 @@
...
+            Self::Utf16CodeUnit => write!(f, "Utf16CodeUnit"),
```
heckj:
```suggestion
            Self::Utf16CodeUnit => write!(f, "UTF-16CodeUnit"),
```

rust/src/text_encoding.rs
```diff
@@ -0,0 +1,42 @@
...
+            am::TextEncoding::Utf16CodeUnit => TextEncoding::Utf16CodeUnit,
+            am::TextEncoding::GraphemeCluster => TextEncoding::GraphemeCluster,
+        }
+    }
+}
+
+impl std::fmt::Display for TextEncoding {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::UnicodeCodePoint => write!(f, "UnicodeCodePoint"),
```
heckj:
```suggestion
            Self::UnicodeCodePoint => write!(f, "UnicodeScalarCodePoint"),
```

heckj: I thought in swift this was represented by `.unicodeScalar` - please correct me if I'm off, as I don't know the text representations all that well. If so, I'd like to keep "scalar" in the visual name for clarity.

rust/src/text_encoding.rs
heckj: As long as hyphens are not invalid here, I'd like to keep to the name format you'd see in written text: `utf-8` or `UTF-8`.

rust/src/text_encoding.rs
heckj: As long as hyphens are not invalid here, I'd like to keep to the name format you'd see in written text: `utf-16` or `UTF-16`

heckj (changes requested):
I've requested a few minor nits/changes in documentation wording for clarity, all as suggested changes. Additionally, I'd like to make sure that the exposed TextEncoding enum is documented through to the external Swift API that it's clear how those values related to Swift encodings.

The tests are the real gem of this PR - thank you - illustrating nicely how the text encoding changes effect cursor positions, etc. 

What, exactly, is this holding on before we can/should merge this? Is there an upstream PR that's pending, or is that yet to be established?

rust/src/text_encoding.rs
miguelangel-dev: addressed in e7d18e95a1eb6b07b55efd35046ed759004c8172

rust/src/text_encoding.rs
miguelangel-dev: addressed in https://github.com/automerge/automerge-swift/commit/e7d18e95a1eb6b07b55efd35046ed759004c8172

rust/src/text_encoding.rs
miguelangel-dev: addressed in https://github.com/automerge/automerge-swift/commit/e7d18e95a1eb6b07b55efd35046ed759004c8172

For our particular use, Unicode scalars are the Unicode code point subset we have implemented in automerge. We delegate it to `rust::char` [[ref](https://github.com/automerge/automerge/blob/a3185700b3400d9f00060ee1dcaeb8675cbab1db/rust/automerge/src/transaction/inner.rs#L601)] which represents a scalar, not a general code point. So, we can proceed with your text representation suggestion.

miguelangel-dev:
> I've requested a few minor nits/changes in documentation wording for clarity, all as suggested changes. Additionally, I'd like to make sure that the exposed TextEncoding enum is documented through to the external Swift API so that it's clear how those values relate to Swift encodings.

Good call. Added Swift type and API documentation about `TextEncoding` in 71d44085f334f08124c3ace028c486509567b574

> What, exactly, is this holding on before we can/should merge this? Is there an upstream PR that's pending, or is that yet to be established?

**Exactly**. This PR exposes changes done in upstream PR related to [TextEncoding](https://github.com/automerge/automerge/pull/997), ~and [Cursor](https://github.com/automerge/automerge/pull/994) fixes~. @alexjg is working on merging and releasing a new automerge version.

rust/src/text_encoding.rs
heckj: Awesome, thank you for that update @miguelangel-dev !

heckj (approved):

miguelangel-dev:
@heckj requested your review because I changed the `ci`. Wasm job uses locally compiled `libuniffi_automerge`, but in mac, we were using the latest release precompiled Automerge. That change will align both and it opens to have two versions of the Automerge library: develop | release - in another way, in develop bindings won't be able to run ci using the latest Automerge library.

**side note:** PR is ready to merge.

heckj (approved):

PR #207, opened 2025-01-28
Support cursors at end of document

miguelangel-dev:
Cursors API tracks the following character; it is the expected result in collaboration for TextEditors. Because of that, Automerge Cursors API, is not expected to accept end of document (length) as valid input. It is expected to track existing characters. It is fine, except by selection. In cursor selection, [0, length] is the valid range. Because the end of the document will define the future inputs of the user.

In that [upstream PR](https://github.com/automerge/automerge/pull/994), Automerge introduced special cases to support exactly that behaviour. Instead of introducing an alternative method into Swift bindings, I consider it an implementation detail. Cursors will be able to accept `length` as a valid range, and internally it will be mapped to introduced [CursorPosition](https://docs.rs/automerge/latest/automerge/enum.CursorPosition.html)

### DIFF
--- a/Sources/Automerge/Cursor.swift
+++ b/Sources/Automerge/Cursor.swift
@@ -5,10 +5,10 @@ typealias FfiPosition = AutomergeUniffi.Position
 /// A opaque type that represents a location within an array or text object that adjusts with insertions and deletes to
 /// maintain its relative position.
 ///
-/// Set a cursor using ``Document/cursor(obj:position:)``, or ``Document/cursorAt(obj:position:heads:)`` to place a
+/// Set a cursor using ``Document/cursor(obj:position:)``, or ``Document/cursor(obj:position:heads:)`` to place a
 /// cursor at a previous point in time.
-/// Retrieve the cursor position from the document using ``Document/cursorPosition(obj:cursor:)``, or use
-/// ``Document/cursorPositionAt(obj:cursor:heads:)`` to get the cursor position at a previous point in time.
+/// Retrieve the cursor position from the document using ``Document/position(obj:cursor:)``, or use
+/// ``Document/position(obj:cursor:heads:)`` to get the cursor position at a previous point in time.
 public struct Cursor: Equatable, Hashable, Sendable {
     var bytes: [UInt8]
 }
@@ -24,7 +24,7 @@ extension Cursor: CustomStringConvertible {
 ///
 /// ### See Also
 /// - ``Document/cursor(obj:position:)``
-/// - ``Document/cursorAt(obj:position:heads:)``
+/// - ``Document/cursor(obj:position:heads:)``
 public enum Position {
     case cursor(Cursor)
     case index(UInt64)
--- a/Sources/Automerge/Cursor.swift
+++ b/Sources/Automerge/Cursor.swift
@@ -2,13 +2,13 @@ import enum AutomergeUniffi.Position
 
 typealias FfiPosition = AutomergeUniffi.Position
 
-/// A opaque type that represents a location within an array or text object that adjusts with insertions and deletes to
+/// A opaque type that represents a stable location of the character following the location reference at creation within an array or text object that adjusts with insertions and deletions to
 /// maintain its relative position.
 ///
-/// Set a cursor using ``Document/cursor(obj:position:)``, or ``Document/cursor(obj:position:heads:)`` to place a
-/// cursor at a previous point in time.
-/// Retrieve the cursor position from the document using ``Document/position(obj:cursor:)``, or use
-/// ``Document/position(obj:cursor:heads:)`` to get the cursor position at a previous point in time.
+/// Create a cursor using ``Document/cursor(obj:position:)``, or ``Document/cursor(obj:position:heads:)`` to place a
+/// cursor at the point in time indicated by the `heads` parameter.
+/// Retrieve the position of the cursor reference from the document using ``Document/position(obj:cursor:)``, or use
+/// ``Document/position(obj:cursor:heads:)`` to get the position at a previous point in time.
 public struct Cursor: Equatable, Hashable, Sendable {
     var bytes: [UInt8]
 }
--- a/Sources/Automerge/Document.swift
+++ b/Sources/Automerge/Document.swift
@@ -583,28 +583,28 @@ public final class Document: @unchecked Sendable {
     /// Establish a cursor at the position you specify in the list or text object you provide.
     ///
     /// In collaborative applications, maintaining stable cursor positions is crucial. Traditional index-based
-    /// positions can become outdated due to document modifications. This method ensures the cursor stays
+    /// positions become outdated when you or collaborators modify the document. This method provides a stable reference to a character that stays
     /// correctly anchored regardless of changes.
     ///
-    /// `Cursor` provides a reliable way to track positions over time without being affected by document changes.
+    /// `Cursor` provides a reliable way to track the position of a character's location over time regardless of document changes.
     /// The cursor remains anchored to the following character, and if placed at the end of the document,
-    /// it will persistently stay attached to the end.
-    ///
+    /// the location it represents stays persistently at the end of the text or array.
+    /// The following snippet provides examples of cursor locations and the character it tracks:
     /// ```swift
     /// "ABC"   // scenario
     /// "A|BC"  // set cursor at `1`, cursor is attached to `B`
     /// "AZ|BC" // insert `Z` at `1`
     /// ```
     ///
-    /// To retrieve the original absolute index-based positions, use:
+    /// To retrieve the absolute index-based positions for a cursor, use:
     /// - ``position(obj:cursor:)``
     /// - ``position(obj:cursor:heads:)``
     ///
     /// - Parameters:
     ///   - obj: The object identifier of the list or text object.
     ///   - position: The index position in the list, or index for a text object based on ``TextEncoding``.
-    ///     When using a position equal to or greater than the current length of the object,
-    ///     the cursor will track the end of the document as it changes.
+    ///     When you use a position equal to or greater than the current length of the object,
+    ///     the cursor tracks the end of the document, regardless of changes.
     /// - Returns: A cursor that references the position you specified.
     ///
     /// ### See Also
@@ -663,8 +663,7 @@ public final class Document: @unchecked Sendable {
 
     /// Retrieves the absolute index-based position for the list or text object you provide.
     ///
-    /// While cursors provide stable positions in a collaborative environment, this method allows you to convert
-    /// a cursor back into an absolute index-based position.
+    /// Use this method to convert a cursor into an absolute index-based position.
     ///
     /// - Parameters:
     ///   - obj: The object identifier of the list or text object.
@@ -685,8 +684,7 @@ public final class Document: @unchecked Sendable {
 
     /// Retrieves the absolute index-based position for the list or text object and point in time you provide.
     ///
-    /// While cursors provide stable positions in a collaborative environment, this method allows you to convert
-    /// a cursor back into an absolute index-based position.
+    /// Use this method to convert a cursor into an absolute index-based position at a previous point in time.
     ///
     /// - Parameters:
     ///   - obj: The object identifier of the list or text object.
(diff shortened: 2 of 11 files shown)


heckj:
Thanks @miguelangel-dev - back me up a second and help me understand the new functionality that's being exposed. 

With this PR we're removing `cursor()` & `cursorAt()`, replacing that (if I'm reading correctly with `cursorSelection` and `cursorSelectionAt` - and then adding a `cursorPosition` (and `cursorPositionAt`). Not that the earlier names were fantastic, but the new names aren't telling me what's going on or implying any clear path of how to use these, and how and what a Cursor object represents when used in a Swift context. 

I'm guessing from the parameters that the first is returning a cursor reference, and the second (new cursorPosition) takes the cursor and returns back an index position from it - so a sort of inverse. Are the `Cursor` types returned expected to be ephemeral, or persistent - a sort of identifier stored in the Automerge document. Can you stash this in some value/encoded form, and then reference again later, for example - when an App opens up, re-hydrating the Cursor, and then it still references a position?

Are there any mechanisms for moving the cursor around, or is this just a "position" that shifts around as document updates get applied?

@alexjg feel free to chime in here - this is just a portion of the API that I haven't really explored and used, and the use cases around it aren't super clear to me.

My only concern here is that the naming of the API  presented seems really unclear, and just from reading them I wouldn't have any sense of what and how I should use them.

miguelangel-dev:
@heckj thanks for taking the time to review the PR:

> With this PR we're removing `cursor()` & `cursorAt()`, replacing that (if I'm reading correctly with `cursorSelection` and `cursorSelectionAt`- and then adding a`cursorPosition`(and`cursorPositionAt`)

Thats it. PR is not modifying existing `position` methods. Just introducing a new case into `cursor` for supporting `length` as a valid input to set a cursor to the end of the document. Previously, accepted range was `[0, length - 1]`

> When an App opens up, re-hydrating the Cursor, and then it still references a position?...  Are there any mechanisms for moving the cursor around, or is this just a "position" that shifts around as document updates get applied?

I agree that naming is not the best one, and probably the documentation neither. In 872a39dbf50bee205ab184056e9af25424c5f5e4 you will notice I have made the public API more swiftly and introduced some clarifications in the documentation.

Using absolute index-based positions in a text document is unreliable, especially in collaborative environments, since any document modification can disrupt indices, requiring additional tracking. Instead, `Cursor` provides a reliable way to track positions over time without being affected by document changes. The cursor remains anchored to the following character. However, in some cases, retrieving the absolute index may be necessary for temporary calculations, such as updating user selections. This is where the inverse operation `position` comes into play.

alexjg:
The use of "cursor" here is by analogy to it's use in the database world as a stable reference to some position in a sequence. This may actually be quite a confusing name because it's a bit different to a cursor in the world of text editing. The main difference is that the reference we return here points at a specific element of a sequence whereas a text editing cursor points at the gap between elements of the sequence (the characters).

The references returned by the cursor API are a bit like object IDs in that they are stable identifiers which can be stored or transmitted and later used to resolve to a position in the sequence they were created for. `cursorPosition` should probably be named something more evocative like `resolveCursor`.

I would be open to changing the name of the type entirely. At some point we will generalise this concept to be able to point to any part of the document so maybe a good name would be something like "stable reference".

miguelangel-dev:
I am open to refactoring the names `cursor` and `position` in public API to whatever aligns well with future iterations in Automerge. Or to keep the consistency with the rest of the existing API. Just let me know what is preferable. 

On my side, using `cursor(at:)` to refer to the logical reference and `position(for:)` to resolve the absolute index aligns well in general with the text editor's environment.

Sources/Automerge/Cursor.swift
```diff
@@ -5,10 +5,10 @@ typealias FfiPosition = AutomergeUniffi.Position
 /// A opaque type that represents a location within an array or text object that adjusts with insertions and deletes to
```
heckj:
```suggestion
/// A opaque type that represents a stable location of the character following the location reference at creation within an array or text object that adjusts with insertions and deletions to
```

Sources/Automerge/Cursor.swift
```diff
@@ -5,10 +5,10 @@ typealias FfiPosition = AutomergeUniffi.Position
...
 /// maintain its relative position.
 ///
-/// Set a cursor using ``Document/cursor(obj:position:)``, or ``Document/cursorAt(obj:position:heads:)`` to place a
+/// Set a cursor using ``Document/cursor(obj:position:)``, or ``Document/cursor(obj:position:heads:)`` to place a
```
heckj:
```suggestion
/// Create a cursor using ``Document/cursor(obj:position:)``, or ``Document/cursor(obj:position:heads:)`` to place a
```

Sources/Automerge/Cursor.swift
```diff
@@ -5,10 +5,10 @@ typealias FfiPosition = AutomergeUniffi.Position
...
 /// cursor at a previous point in time.
```
heckj:
```suggestion
/// cursor at the point in time indicated by the `heads` parameter.
```

Sources/Automerge/Cursor.swift
```diff
@@ -5,10 +5,10 @@ typealias FfiPosition = AutomergeUniffi.Position
...
-/// Retrieve the cursor position from the document using ``Document/cursorPosition(obj:cursor:)``, or use
-/// ``Document/cursorPositionAt(obj:cursor:heads:)`` to get the cursor position at a previous point in time.
+/// Retrieve the cursor position from the document using ``Document/position(obj:cursor:)``, or use
```
heckj:
```suggestion
/// Retrieve the position of the cursor reference from the document using ``Document/position(obj:cursor:)``, or use
```

Sources/Automerge/Cursor.swift
```diff
@@ -5,10 +5,10 @@ typealias FfiPosition = AutomergeUniffi.Position
...
+/// ``Document/position(obj:cursor:heads:)`` to get the cursor position at a previous point in time.
```
heckj:
```suggestion
/// ``Document/position(obj:cursor:heads:)`` to get the position at a previous point in time.
```

Sources/Automerge/Document.swift
```diff
@@ -582,10 +582,34 @@ public final class Document: @unchecked Sendable {
 
     /// Establish a cursor at the position you specify in the list or text object you provide.
     ///
+    /// In collaborative applications, maintaining stable cursor positions is crucial. Traditional index-based
+    /// positions can become outdated due to document modifications. This method ensures the cursor stays
```
```suggestion
    /// positions become outdated when you or collaborators modify the document. This method provides a stable reference to a character that stays
```

Sources/Automerge/Document.swift
```diff
@@ -582,10 +582,34 @@ public final class Document: @unchecked Sendable {
...
+    /// correctly anchored regardless of changes.
+    ///
+    /// `Cursor` provides a reliable way to track positions over time without being affected by document changes.
```
```suggestion
    /// `Cursor` provides a reliable way to track the position of a character's location over time regardless of document changes.
```

Sources/Automerge/Document.swift
```diff
@@ -582,10 +582,34 @@ public final class Document: @unchecked Sendable {
...
+    /// The cursor remains anchored to the following character, and if placed at the end of the document,
+    /// it will persistently stay attached to the end.
```
```suggestion
    /// the location it represents stays persistently at the end of the text or array.
```

Sources/Automerge/Document.swift
```diff
@@ -582,10 +582,34 @@ public final class Document: @unchecked Sendable {
...
+    ///
```
```suggestion
    /// The following snippet provides examples of cursor locations and the character it tracks:
```

Sources/Automerge/Document.swift
```diff
@@ -582,10 +582,34 @@ public final class Document: @unchecked Sendable {
...
+    /// ```swift
+    /// "ABC"   // scenario
+    /// "A|BC"  // set cursor at `1`, cursor is attached to `B`
+    /// "AZ|BC" // insert `Z` at `1`
+    /// ```
+    ///
+    /// To retrieve the original absolute index-based positions, use:
```
```suggestion
    /// To retrieve the absolute index-based positions for a cursor, use:
```

Sources/Automerge/Document.swift
```diff
@@ -582,10 +582,34 @@ public final class Document: @unchecked Sendable {
...
+    /// - ``position(obj:cursor:)``
+    /// - ``position(obj:cursor:heads:)``
+    ///
     /// - Parameters:
     ///   - obj: The object identifier of the list or text object.
     ///   - position: The index position in the list, or index for a text object based on ``TextEncoding``.
+    ///     When using a position equal to or greater than the current length of the object,
```
```suggestion
    ///     When you use a position equal to or greater than the current length of the object,
```

Sources/Automerge/Document.swift
```diff
@@ -582,10 +582,34 @@ public final class Document: @unchecked Sendable {
...
+    ///     the cursor will track the end of the document as it changes.
```
```suggestion
    ///     the cursor tracks the end of the document, regardless of changes.
```

Sources/Automerge/Document.swift
```diff
@@ -596,12 +620,36 @@ public final class Document: @unchecked Sendable {
 
     /// Establish a cursor at the position and point of time you specify in the list or text object you provide.
     ///
+    /// In collaborative applications, maintaining stable cursor positions is crucial. Traditional index-based
```
heckj: ^^ I'd like to see the same grammar and language tweaks in this documentation set as well. The changes are mostly aimed at removing passive voice and being explicit about what's doing the changing or tracking.

Sources/Automerge/Document.swift
```diff
@@ -613,28 +661,44 @@ public final class Document: @unchecked Sendable {
         }
     }
 
-    /// The current position of the cursor for the list or text object you provide.
+    /// Retrieves the absolute index-based position for the list or text object you provide.
+    ///
+    /// While cursors provide stable positions in a collaborative environment, this method allows you to convert
```
```suggestion
    /// Use this method to convert a cursor into an absolute index-based position.
```

Sources/Automerge/Document.swift
```diff
@@ -613,28 +661,44 @@ public final class Document: @unchecked Sendable {
...
+    /// a cursor back into an absolute index-based position.
```
```suggestion
```

Sources/Automerge/Document.swift
```diff
@@ -613,28 +661,44 @@ public final class Document: @unchecked Sendable {
...
             try self.doc.wrapErrors {
                 try $0.cursorPosition(obj: obj.bytes, cursor: cursor.bytes)
             }
         }
     }
 
-    /// The historical position of the cursor for the list or text object and point in time you provide.
+    /// Retrieves the absolute index-based position for the list or text object and point in time you provide.
+    ///
+    /// While cursors provide stable positions in a collaborative environment, this method allows you to convert
```
```suggestion
    /// Use this method to convert a cursor into an absolute index-based position at a previous point in time.
```

Sources/Automerge/Document.swift
```diff
@@ -613,28 +661,44 @@ public final class Document: @unchecked Sendable {
...
+    /// a cursor back into an absolute index-based position.
```
```suggestion
```

heckj (commented):
@miguelangel-dev sorry for the delay! 

And thank you both for chiming in to help me understand. I've asked for a few changes to the documentation, and I'm happy with the updated names. I'd guess using "cursor" will still run afoul of someone being confused, as it's a fairly overloaded term, but with including the detail that it tracks a specific *character* through changes should help alleviate that - at least for the folks who read the docs.

I couldn't come up with any better names myself, and I think `cursor` and it's inverse of `position` work well for what this API is doing. Again, thank you for shortening to the simpler, more direct names - I think that'll help tremendously.

heckj (changes requested):
(I'd submitted the last review as "comment" when I meant to mark is at "requesting changes")

Sources/Automerge/Cursor.swift
miguelangel-dev: Ohh good one!

Sources/Automerge/Document.swift
miguelangel-dev: Indeed, your recent grammar tweaks have rendered it redundant, so we can remove it for clarity.  Addressed in: f7657f61896f70ca1d6dc36b74355274daf20b8a

Sources/Automerge/Document.swift
miguelangel-dev: Indeed, your recent grammar tweaks have rendered it redundant, so we can remove it for clarity. Addressed in: https://github.com/automerge/automerge-swift/commit/f7657f61896f70ca1d6dc36b74355274daf20b8a

miguelangel-dev:
> I couldn't come up with any better names myself, and I think `cursor` and it's inverse of `position` work well for what this API is doing. Again, thank you for shortening to the simpler, more direct names - I think that'll help tremendously.

I've implemented your suggestions—thanks for the help with the grammar and readability improvements. The documentation now has a much more polished ❤️

heckj (approved):

PR #220, opened 2025-11-19
feat: update automerge in rs to 0.7.1

situ2001:
I am not sure this update may break other things, since it is a coordinated update accompanying the upgrade of automerge-js from v2 to v3, but all Swift test are passed.

What I did is:

1. update the version of rust deps.
2. Run `cargo check` to detect all the errors and fix them.
3. Run `swift test`.


lightsprint09:
Thanks for bringing this up. I am not to familiar with the rust-ffi, but have you recompiled automerge-rust. If not the tests still use the old xcframework to run the tests

lightsprint09:
Basend on your work I completed the release. See https://github.com/automerge/automerge-swift/pull/221

